PR #12, opened 2024-10-19
Add back submodules and ci

xuzhao9:
submodules and ci were mistakenly deleted during setting up internal sync.
Re-adding them to resume the sync.


adamomainz (approved):
Thanks for adding this back in!

PR #48, opened 2024-11-13
Add an init file to tools so that it is considered as an module.

htyu:
Fixing issues like:

  File "tritonbench/install.py", line 8, in <module>
    from tools.cuda_utils import CUDA_VERSION_MAP, DEFAULT_CUDA_VERSION
ModuleNotFoundError: No module named 'tools.cuda_utils'


xuzhao9 (approved):
LGTM!

PR #56, opened 2024-11-19
Enable gemm and more operators in the CI

xuzhao9:
As the test isolation is implemented in https://github.com/pytorch-labs/tritonbench/pull/55, we can now enable more operators in the CI.


adamomainz (approved):

adamomainz:
Nice job!

PR #74, opened 2024-11-22
Enable bwd for flash_attention

xuzhao9:
Some backends (e.g., flash_attention/xformers_splitk) don't have backward pass. For those backends, add `fwd_only=True` flag, and we will skip the backward pass automatically.

If user specifies `--only xformers_splitk --bwd`, we will still run the backend since it is user-specified. Otherwise we will always skip this backend.

Test plan:

```
python -m unittest test.test_gpu.main -k flash_attention
```

### DIFF
--- a/test/test_gpu/main.py
+++ b/test/test_gpu/main.py
@@ -43,11 +43,16 @@ TEST_OPERATORS = set(list_operators_by_collection(op_collection="default"))
 
 
 def check_ci_output(op):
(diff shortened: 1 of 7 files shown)


adamomainz:
can you please share a screenshot of the output or confirm it ran as expected?

xuzhao9:
@adamomainz   Sorry it is not ready yet and I am still debugging - I will request review again when it is ready

adamomainz (approved):

PR #81, opened 2024-11-26
Install tritonbench as a library

xuzhao9:
Now users can import Tritonbench as a library:

```
$ pip install -e .
$ python -c "import tritonbench; op = tritonbench.load_opbench_by_name('addmm');"
```

Clean up the init file so that dependencies like `os` and `importlib` will not pollute the namespace.


FindHao (approved):
LGTM!

PR #91, opened 2024-12-03
[FA] add bwd variants for warp spec

manman-ren:


xuzhao9:
We have fixed the CI on main, can you run rebase to re-trigger the CI tests?

test/test_gpu/skip_tests_h100_pytorch.yaml
```diff
@@ -19,6 +19,7 @@ flash_attention:
   - triton_tutorial_flash_v2_ws
   - triton_tutorial_flash_v2_tma_ws
   - triton_tutorial_flash_v2_tma_ws_persistent
+  - triton_tutorial_flash_v2_bwd_ws
```
xuzhao9, line 22: We do not have a separate name for bwd kernel in the skiplist. Since `triton_tutorial_flash_v2_ws` has been included, there is no need to add `triton_tutorial_flash_v2_bwd_ws`.

manman-ren: triton_tutorial_flash_v2_bwd_ws is a different benchmark from triton_tutorial_flash_v2_ws:
```
    @register_benchmark()
    def triton_tutorial_flash_v2_bwd_ws( ...
        return lambda: triton_tutorial_FA2_opt(
            q, k, v, self.causal, self.sm_scale, "base", "ws"
        )
```
Do I need to include it in the skiplist?

xuzhao9: Ah I see, yes, we need to include it in the skip list in this case.

Just asking, what does "bwd" mean here? Does it mean backwards?

manman-ren: Yes it means backwards. I added two extra arguments to specify variants for fwd and bwd.

xuzhao9: I am wondering why we need to add explicit benchmark for backward? I thought it could be covered by calling `--mode bwd` and `get_bwd_fn()`: https://github.com/pytorch-labs/tritonbench/blob/main/tritonbench/operators/flash_attention/operator.py#L484

manman-ren: Yeah, this diff does not add explicit benchmark for backward, we still use --bwd to run the backward kernel, but there are two implementations for the backward kernel and I used a different benchmark name to choose between the two variants. Open to other suggestions :]
Maybe we can add a command line arg, but I want to compare the accuracy between the two implementations.

xuzhao9: Ah I see, then creating a separate benchmark sounds good to me.

xuzhao9 (approved):
LGTM!

We may later add a flag to `@register_benchmark(bwd_only=True)` for the kernels that are only meant to run in backwards.

PR #104, opened 2024-12-07
Disable donated_buffer for all ops's backward benchmarking

FindHao:
It is still a temporary fix for backward benchmarking. Related discussion https://github.com/pytorch-labs/tritonbench/issues/40

### DIFF
--- a/tritonbench/utils/triton_op.py
+++ b/tritonbench/utils/triton_op.py
@@ -623,6 +623,7 @@ class BenchmarkOperator(metaclass=PostInitProcessor):
         if self.mode in [Mode.FWD_BWD, Mode.BWD]:
             # TODO: remove this once we have a better way to handle backward benchmarking
             import torch._functorch.config
+
             torch._functorch.config.donated_buffer = False
         self.device = tb_args.device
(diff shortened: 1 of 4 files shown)


tritonbench/utils/triton_op.py
```diff
@@ -39,6 +39,8 @@
     tqdm = None
 
 logger = logging.getLogger(__name__)
+# TODO: remove this once we have a better way to handle backward benchmarking
+torch._functorch.config.donated_buffer = False
```
xuzhao9: This seems an overkill to me. Can we disable it only in backward and forward_backward?

FindHao: fixed in https://github.com/pytorch-labs/tritonbench/pull/104/commits/20b138d024ab39697642fb6bf9db8c9d969ebd5e

xuzhao9 (approved):

FindHao:
It looks the ci errors are true, and the config is not compatible with flash_attention. I'm going to set this config for `fused_linear_cross_entropy, geglu, and swiglu ` only.

PR #106, opened 2024-12-07
[metrics] Enable cudagraph mode for kineto_trace

xuzhao9:
- Remove `use_cuda_graph = False` as we are now using non cudagraph as default. Many operators do not support cudagraph mode.
- Enable `--cudagraph --metrics kineto_trace` to turn on cuda graph mode for Kineto trace.


FindHao (approved):
LGTM! Thanks for the updates!

PR #111, opened 2024-12-11
Support flops metric in proton profiling

xuzhao9:
Summary:
Support tflops+proton metric on more operators.

Now we require operator to use `def flops(): ` and the framework will add `tflops` metric automatically.
The same interface will be used to support `flops` in Proton.

Differential Revision: D67062546


FindHao (approved):
LGTM!

PR #120, opened 2024-12-19
[docker] Fix the nightly docker build

xuzhao9:
After the submodule update, the FA3 CUTLASS kernels now cost much more memory to build and will exhaust the CI machine.

Make the following changes:
1. Build FA3 with MAX_JOBS=4 and NVCC_THREADS=1.
2. Disable xformers build as it also requires FA3, we will try to enable its build later.
3. Also disable colfax and TK build for now - will fix that later.

Test plan: CI


xuzhao9:
Show disk space: https://github.com/pytorch-labs/tritonbench/actions/runs/12421954579

PR #124, opened 2025-01-02
[CI] Fix the CI failure by skipping xformers

xuzhao9:
We have decided to skip xformers compilation in the nightly docker build because of limited docker compilation time.
Therefore we will not run or test xformers fa kernels in the CI. The tests are still available internally.


xuzhao9:
The main branch test is still broken, because Triton main is incompatible with TorchInductor at this point.

PR #125, opened 2025-01-06
Add reset dynamo option

FindHao:
Fix https://github.com/pytorch-labs/tritonbench/issues/90

When we run multiple inputs at one time, it may let torch._dynamo hit config.cache_size_limit. This PR adds an option to reset dynamo for an operator.


xuzhao9 (approved):
LGTM

PR #142, opened 2025-01-24
[tritonbench] Change default precision for addmm to fp16

SamGinzburg:
streamk was recently added to the addmm benchmarks---but atomic_add does not support bf16 in upstream Triton, so we should change the default to fp16.


xuzhao9 (approved):
LGTM

PR #148, opened 2025-01-30
[compile_trace] Add compile time Kineto trace

xuzhao9:
Generates Kineto trace focusing on compile time so that we can peek into compile time analysis.

Fixes https://github.com/pytorch-labs/tritonbench/issues/117

Test plan:

```
$ python run.py --op softmax --num-inputs 1 --input-id 0 --metrics compile_trace --only triton_softmax
```

image

Internal link only: https://www.internalfb.com/intern/perfdoctor/trace_view?filepath=tree%2Ftritonbench%2Fcompile_time.json&bucket=tc_bench_ci

With autotuning:

image

### DIFF
--- a/tritonbench/components/compile_time/__init__.py
+++ b/tritonbench/components/compile_time/__init__.py
@@ -1,1 +1,5 @@
-from .trace import do_compile_time_in_task, fbcode_do_compile_time_in_task  # noqa F401
+from .trace import (  # noqa F401
+    do_compile_kineto_trace_in_task,
+    do_compile_time_in_task,
+    fbcode_do_compile_time_in_task,
+)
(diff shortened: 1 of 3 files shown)


xuzhao9:
cc @fywkevin @Jokeren as it is a metrics improvement, it should be a Python view of the Triton compilation process, but Kineto has no observation to the C++ code.

xuzhao9:
For autotuning, the Triton's autotuner has some GPU idleness gaps:

Jokeren:
> This is a Python side view of the Triton compilation process, note that Kineto has no observation on the C++ code.

I think it's fine to have such a coarse-grained view from profilers. 
MLIR can get you a very detailed view of time spent on each compiler passe, but I think it may only provide insights for compiler engineers.

PR #153, opened 2025-02-04
[compile_time] Deploy compile_time benchmark CI

xuzhao9:
As the title says. It will upload the result to pytorch CI AWS S3 bucket and ready to query in ClickHouse.

Test plan:

https://github.com/pytorch-labs/tritonbench/actions/runs/13205478992/job/36867552165?pr=153


.github/workflows/_linux-benchmark-h100.yml
```diff
@@ -27,8 +35,11 @@ jobs:
     env:
```
huydhn, line 38: I think you need to set `id-token` here too:

```
permissions:
  id-token: write
  contents: read
```

xuzhao9: Thanks! Yes it fixes the problem.

PR #165, opened 2025-02-19
Fix embedding_dim for fp8 attention

mandroid6:
Summary: By default we set `embedding_dim` to be 3072, but its not expected when ad_head` is also provided as an input to the benchmark resulting in a value error.

Differential Revision: D69881611


xuzhao9:
Superceded by https://github.com/pytorch-labs/tritonbench/pull/139

PR #206, opened 2025-04-22
Remove make_block_ptr from Attention forward

njriasan:
Summary:
Removes make_block_ptr from HSTU forward as its deprecated and prevents BufferOps on AMD. A followup diff with enable bufferops.

This diff functions as a simplified trial run to enabling BufferOps in Hammer's HSTU. Since this implementation is much simpler it should be less likely to have any errors.

Differential Revision: D73382705


xuzhao9 (approved):
LGTM

PR #211, opened 2025-04-24
add triton operators for nightly

xuzhao9:
We use the following criteria to adopt a nightly benchmark operator:

1. Must have at least one Triton backend
2. Either supports the tflops metric, OR
3. If it has a non-Triton baseline backend, calculate the speedup comparing with the baseline.

* __->__ #211
* #210


xuzhao9:
Need to run on fix install: https://github.com/pytorch-labs/tritonbench/actions/runs/14648182559

PR #219, opened 2025-05-08
[submodules] Update cutlass and other submodules

xuzhao9:
As the title says.

For cutlass, we need to be consistent with pytorch upstream: https://github.com/pytorch/pytorch/tree/main/third_party


xuzhao9:
Found upstream issue: https://github.com/meta-recsys/generative-recommenders/issues/276

PR #251, opened 2025-06-25
[ci][nightly] Enable more tests

xuzhao9:
BE efforts enables more tests in nightly.

Note that we can't enable a couple of fp8_* operators from FBGEMM, because FBGEMM Triton code still uses the old TMA API and therefore can't work on Triton main.

Test plan:

https://github.com/pytorch-labs/tritonbench/actions/runs/16037334650/job/45251972846?pr=251

### DIFF
--- a/benchmarks/nightly/autogen.yaml
+++ b/benchmarks/nightly/autogen.yaml
@@ -8,31 +8,28 @@ cross_entropy_fwd:
 embedding_fwd:
   op: embedding
   args: --op embedding --baseline torch_embedding --metrics speedup --only liger_embedding,torch_embedding
-# TODO: fix in the CI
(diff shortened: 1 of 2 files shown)


davidberard98 (approved):
quick question, is `autogen.yaml` auto-generated? I was imagining we shouldn't manually modify autogen.yaml based on its name

xuzhao9:
> quick question, is `autogen.yaml` auto-generated? I was imagining we shouldn't manually modify autogen.yaml based on its name

It is generated by `https://github.com/pytorch-labs/tritonbench/blob/xz9/add-nightly/benchmarks/nightly/gen.py`, it is fine for now, I will submit a follow-up PR to update the generator script.

PR #259, opened 2025-07-02
Support old triton versions w/ setup.py in ./python

davidberard98:
I want to ab test triton 3.3 and triton 3.4 to look for perf regressions or failures not caught in PyTorch CI. But Triton 3.3 is too old for the existing setup script, which assumes setup.py in the ./ directory (in triton 3.3, it's in ./python/setup.py).

This PR adds support to build either version.

before (fails to compile 3.3): https://github.com/pytorch-labs/tritonbench/actions/runs/16012761341
after (succeeds in compiling 3.3): https://github.com/pytorch-labs/tritonbench/actions/runs/16013163616


xuzhao9 (approved):
LGTM!

PR #260, opened 2025-07-02
[FA] Add code to enable autoWS from triton-lang

manman-ren:
Also clean up setting up configs for different optimization variants.
- Extract computation in inner loop as a helper function attn_fwd_iteration
- We have 3 different versions of attn_fwd_inner: attn_fwd_inner[_autows|_ws], _autows is for OSS triton where we use warp_specialize on ForOp to enable warpspec, _ws is for internal triton, where we use tl.async_task and num_consumer_groups to enable warpspec
- We have 2 different versions of attn_fwd_compute: attn_fwd_compute[_ws], attn_fwd_compute_ws calls attn_fwd_inner_ws, attn_fwd_compute calls either attn_fwd_inner or attn_fwd_inner_autows depending on whether warpspec should be enabled via argument WARP_SPECIALIZE
- We have 4 top-level kernels: _attn_fwd_ws, _attn_fwd_base_opt, _attn_fwd_tma_unified, _attn_fwd_tma_ws_persistent
- Depending on env var ENABLE_EXPLICIT_WS, _attn_fwd_ws will call either _attn_fwd_compute_ws or attn_fwd_compute with WARP_SPECIALIZE=True

```
TRITON_PRINT_AUTOTUNING=1 WITH_TMA=1 CUDA_VISIBLE_DEVICES=5 TORCH_CUDA_ARCH_LIST=9.0a ./denoise.sh python run.py --op flash_attention --only triton_tutorial_flash_v2_tma_ws_persistent,triton_tutorial_flash_v2_tma_ws --num-inputs 1 --seq-len 8192 --metrics tflops --batch 8 --n-heads 16 --d-head 128
  (Batch, Heads, SeqLen, SeqLen_KV, Dhead)    triton_tutorial_flash_v2_tma_ws_persistent-tflops    triton_tutorial_flash_v2_tma_ws-tflops
------------------------------------------  ---------------------------------------------------  ----------------------------------------
                  (8, 16, 8192, 8192, 128)                                              517.261                                   499.359
```

### DIFF
--- a/tritonbench/kernels/attention_utils.py
+++ b/tritonbench/kernels/attention_utils.py
@@ -15,7 +15,7 @@ HAS_TMA_DESC = "nv_tma_desc_type" in dir(tl)
 WITH_COMPPIPE = os.getenv("ENABLE_COMPPIPE")
 PEEL_LAST = os.getenv("PEEL_LAST_ITER")
 WITH_TMA = os.getenv("WITH_TMA")
-HAS_AUTO_WS = os.getenv("ENABLE_AUTO_WS")
+HAS_EXPLICIT_WS = os.getenv("ENABLE_EXPLICIT_WS")
 
 
 class TmaAutoTuneHelper:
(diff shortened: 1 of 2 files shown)


xuzhao9 (approved):
LGTM, thanks!

PR #262, opened 2025-07-02
fix ab test: "side_b_commit"

davidberard98:
Previously both the "A Side" and "B Side" were running off the same Triton commit - this updates the "B Side" to actually use the side_b_commit instead of the side_a_commit

### DIFF
--- a/.github/workflows/_linux-benchmark-abtest-h100.yml
+++ b/.github/workflows/_linux-benchmark-abtest-h100.yml
@@ -79,7 +79,7 @@ jobs:
           rm -rf .benchmarks || true
       - name: Compile Triton (Side B)
         run: |
-          bash ./.ci/triton/compile.sh --repo ${{ inputs.side_a_triton }} --commit ${{ inputs.side_a_commit }} --side b
+          bash ./.ci/triton/compile.sh --repo ${{ inputs.side_a_triton }} --commit ${{ inputs.side_b_commit }} --side b
       - name: Benchmark Triton (Side B)
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/_linux-benchmark-abtest-h100.yml
```diff
@@ -79,7 +79,7 @@ jobs:
           rm -rf .benchmarks || true
       - name: Compile Triton (Side B)
         run: |
-          bash ./.ci/triton/compile.sh --repo ${{ inputs.side_a_triton }} --commit ${{ inputs.side_a_commit }} --side b
+          bash ./.ci/triton/compile.sh --repo ${{ inputs.side_a_triton }} --commit ${{ inputs.side_b_commit }} --side b
```
xuzhao9: Should this be `--repo ${{ inputs.side_b_triton }}` ?

xuzhao9:
Oops, sorry for the bug, and thanks for the fix!

See my inline comment on whether we should use `${{ inputs.side_b_triton }}` instead.

.github/workflows/_linux-benchmark-abtest-h100.yml
davidberard98:
```suggestion
          bash ./.ci/triton/compile.sh --repo ${{ inputs.side_b_triton }} --commit ${{ inputs.side_b_commit }} --side b
```

davidberard98: thanks for catching!

xuzhao9 (approved):
LGTM!

PR #276, opened 2025-07-11
Add accuracy check and fixes for fp8_attention Triton kernels

yf225:
Stacked PRs:
 * #281
 * __->__#276

--- --- ---

PR #280, opened 2025-07-12
Fix accuracy check for flash_attention kernels

yf225:
Stacked PRs:
 * #281
 * #276
 * __->__#280

--- --- ---

### DIFF
--- a/tritonbench/operators/flash_attention/operator.py
+++ b/tritonbench/operators/flash_attention/operator.py
@@ -201,9 +201,10 @@ class Operator(BenchmarkOperator):
         self.pt2_sdpa = args.pt2_sdpa
         self.additional_inputs = args.additional_inputs
         self.ragged_shapes = args.ragged_shapes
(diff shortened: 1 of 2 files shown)


tritonbench/operators/flash_attention/operator.py
```diff
@@ -524,6 +526,28 @@ def causal_mask(b, h, q_idx, kv_idx):
 
         return lambda: flex_attention(q, k, v, block_mask=block_mask)
 
+    def accuracy(self, fn, baseline_fn):
```
xuzhao9, line 529: cc @manman-ren @y-sq , can we use the same tolerance in blackwell-attention ?

yf225:
@xuzhao9 would you like to also stamp [D78222794](https://www.internalfb.com/diff/D78222794?dst_version_fbid=746939077918945) ? (not sure why the Github stamp is not transferred over..) Thanks so much!

PR #283, opened 2025-07-13
Pass in weight tensor via kernel function arg for fused_linear_cross_entropy

yf225:
Pass in weight tensor via kernel function arg for fused_linear_cross_entropy


tritonbench/operators/fused_linear_cross_entropy/operator.py
```diff
@@ -91,20 +88,20 @@ def get_input_iter(self) -> Generator:
             target = torch.randint(
                 self.vocab_size, (BT, 1), dtype=torch.long, device=self.device
             ).squeeze(1)
-            yield _input, target
+            yield _input, self.weight, target
```
yf225, line 91: This PR effectively "functionalizes" the kernel function, i.e. making it not dependent on external state and instead depends only on its input args. This makes Helion integration easier because the Helion kernel is also stateless.

PR #284, opened 2025-07-13
[wip] set baseline for ragged_attention

yf225:
[wip] set baseline for ragged_attention


xuzhao9 (approved):
It is okay to set `hstu` as baseline in OSS. Internally, we use `hstu_cuda` as the baseline because Triton team is measuring the gap between CUDA and Triton impl.

BTW, users can use `--baseline <BACKEND_NAME>` to specify baseline.

PR #285, opened 2025-07-13
welford: adjust tolorance to make accuracy check pass

yf225:
welford: adjust tolorance to make accuracy check pass


xuzhao9:
Yes we should change all `def _get_accuracy()` (Old API)  to `def accuracy()` (New API) in operator code.

xuzhao9 (approved):

PR #291, opened 2025-07-16
[Fix] Add parser.add_argument for Quack

bdbowyer:
My docker is failing with the following error when it runs install.py.
71.11 File "/app/tritonbench/install.py", line 201, in
71.11 if args.quack or args.all:
71.11 AttributeError: 'Namespace' object has no attribute 'quack'

Added a parser argument for quack.


xuzhao9:
This is replaced by https://github.com/pytorch-labs/tritonbench/pull/293, so closing it.

PR #299, opened 2025-07-22
Add small block shapes to warpspec matmul configs

njriasan:
Summary: Many of the shapes performed much worse, which may be due to the tutorial assuming large configs. This adds smaller configs as an option so we can test warp spec vs just the impact of blackwell configs.

Reviewed By: PaulZhang12

Differential Revision: D78059342


adamomainz (approved):
reviewed and accepted internally

PR #312, opened 2025-07-30
Fix CSV output bug

yf225:
Fix CSV output bug

When `--csv` is specified, we should call `write_csv_to_file` instead of `write_json_to_file`.


njriasan (approved):

yf225:
merged a different version, will close this one

PR #319, opened 2025-07-30
Add --only-match-mode=prefix-with-baseline to enable specific impl prefix  together  with  baseline

yf225:
During Helion development, it's quite common to want to compare Helion impl vs. baseline (since speedup numbers are all compared against baseline). With this PR, we can do `--baseline --only "helion" --only-match-mode=prefix` to enable specific impl prefix to make this much easier to run.

### DIFF
--- a/tritonbench/utils/parser.py
+++ b/tritonbench/utils/parser.py
@@ -129,6 +129,12 @@ def get_parser(args=None):
         help="Specify one or multiple kernel implementations to skip.",
     )
     parser.add_argument(
+        "--only-match-mode",
+        default="exact",
+        choices=["exact", "prefix-with-baseline"],
+        help="Match mode for --only argument. 'exact' for full string match, 'prefix-with-baseline' for prefix match as well as the existing baseline. Default: exact",
+    )
+    parser.add_argument(
         "--baseline", type=str, default=None, help="Override default baseline."
     )
     parser.add_argument(
--- a/tritonbench/utils/triton_op.py
+++ b/tritonbench/utils/triton_op.py
@@ -736,6 +736,7 @@ class BenchmarkOperator(metaclass=PostInitProcessor):
             BASELINE_BENCHMARKS[self.name] = self.tb_args.baseline
         self._only = _split_params_by_comma(self.tb_args.only)
         self._skip = _split_params_by_comma(self.tb_args.skip)
+        self._only_match_mode = self.tb_args.only_match_mode
         self._input_id = self.tb_args.input_id
         self._num_inputs = self.tb_args.num_inputs
         self.prod_shapes = self.tb_args.prod_shapes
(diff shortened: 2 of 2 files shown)


tritonbench/utils/parser.py
```diff
@@ -129,7 +129,16 @@ def get_parser(args=None):
         help="Specify one or multiple kernel implementations to skip.",
     )
     parser.add_argument(
-        "--baseline", type=str, default=None, help="Override default baseline."
+        "--only-prefix",
```
FindHao: Is it possible to update the existing --only for fuzzy matching? or maybe add a `--only-match-mode=prefix` etc.? otherwise, I feel if someone want to add a regex matching in the future, we may need to add another argument for it. what do you think? @adamomainz

tritonbench/utils/triton_op.py
```diff
@@ -858,6 +860,26 @@ def run(
...
+                    if self._only_prefix:
+                        filtered_benchmarks = []
+                        for bm in benchmarks:
+                            for prefix in self._only_prefix:
+                                if bm.startswith(prefix):
+                                    filtered_benchmarks.append(bm)
+                                    break
+                        benchmarks = filtered_benchmarks
+
+                    # Handle --baseline=True flag (i.e. enable existing baseline)
```
FindHao: I don't get it here. by default, if you have set baseline in your op, it should be enabled already, right? why we need to set the command line argument `--baseline`to True?

yf225: We need a way to make `--baseline` and `--only` (or `--only-prefix`) run together. Right now if `--only` is specified, then baseline is not run.

One could argue that "ok then you should also specify the baseline name in the --only", but it would really take an engineer's time to find out the name of the baseline function. It would be good if `--baseline` can just enable the baseline func regardless of other flags.

tritonbench/utils/parser.py
yf225: `--only-match-mode=prefix` sounds good to me too

tritonbench/utils/triton_op.py
FindHao: make sense!

FindHao:
The update plan and clarification sounds good to me. 

Can you:
- add the `--only-match-mode`
- export to internal diff and see if we break something on ci tests?

adamomainz:
This seems really confusing to me. Is this just so you dont have to have a different command per operator? I am not a huge fan of changing the `--baseline` logic further. Right now this flag is not used to add a benchmark to an operator its used to override which operator is denoted as the baseline. Seems we are overloading it a bit here which is where some of the confusion is coming from. Not going to block you landing if @FindHao has already approved but would prefer chatting through before landing.

yf225:
> Is this just so you dont have to have a different command per operator?

@adamomainz yes that’s right - it’s super common to want to compare a prefix of implementations (let’s say “triton” or “Helion”) against the baseline, and needing to dig up the baseline name to pass into —only is not good UX. So would like to look for a solution here.

Worse case, I can add this to the Helion+TB integration code: iterate through all registered impls to see which one is baseline and pass that into —only. But ideally would still prefer having this capability within TB itself.

adamomainz:
> @adamomainz yes that’s right - it’s super common to want to compare a prefix of implementations (let’s say “triton” or “Helion”) against the baseline, and needing to dig up the baseline name to pass into —only is not good UX. So would like to look for a solution here.

@yf225 I understand you. Do you want to just add into the `--only-match-mode` an option for `prefix-with-baseline` or something instead of changing the `--baseline` implementation? I think it would make things clearer and still get what you want?  also makes it easy to extend the match modes if we want more in the future!

adamomainz (approved):
Thanks for making changes based on our discussion so quickly! As long as local testing is passing for you and you are still getting your desired outcomes I am happy to see this land 😃  Please import to fbsource first to make sure all signals pass and then land away!

yf225:
@adamomainz @FindHao I wonder if this is already landed on fbcode side [D79310664](https://www.internalfb.com/diff/D79310664?dst_version_fbid=1277477963789995), do I still need to merge the PR here or it will be done automatically? Thanks!

FindHao:
> @adamomainz @FindHao I wonder if this is already landed on fbcode side [D79310664](https://www.internalfb.com/diff/D79310664?dst_version_fbid=1277477963789995), do I still need to merge the PR here or it will be done automatically? Thanks!

TritonBench now use OSS repo as the first priority via difftrain when sync internal and oss codebase. So, IIUC, we are supposed to merge the OSS PR and then land the internal diff. but I can be wrong. need @xuzhao9 to confirm that.

PR #320, opened 2025-07-31
Add TLX attention (WS pipelined pingpong hopper)

yf225:


tritonbench/operators/flash_attention/operator.py
```diff
@@ -299,6 +311,16 @@ def triton_tutorial_flash_v2_tma(
             q, k, v, self.causal, self.sm_scale, "tma"
         )
 
+    @register_benchmark(enabled=HAS_TLX)
```
njriasan, line 314: Is the intention to run on CI? If you just want to trigger this test directly you can do with `enabled=False` and without any risk of it running on hardware you don't expect (e.g. MI300), which it will by default right now.

tritonbench/operators/flash_attention/tlx_attn_ws_pipelined_pingpong_hopper.py
```diff
@@ -0,0 +1,383 @@
+# Copied from https://github.com/facebookexperimental/triton/blob/tlx/third_party/tlx/tutorials/flash-attention-WS-pipelined-pingpong-hopper.py
+
+import torch
+
+import triton
+import triton.language as tl
+import triton.language.extra.tlx as tlx
```
njriasan, line 7: I'm assuming we have tested this doesn't crash without tlx because of the guards in the operator.py file.

njriasan (commented):
See my comment but otherwise this looks good. Thanks!

tritonbench/operators/flash_attention/operator.py
```diff
@@ -136,6 +136,18 @@
 except (ImportError, IOError, AttributeError, TypeError):
     HAS_XFORMERS = False
 
+# [Optional] TLX backend
+try:
```
xuzhao9, line 140: We introduced a cleaner version of the guards in https://github.com/meta-pytorch/tritonbench/pull/387. This can be simplified to 
```suggestion
from tritonbench.utils.python_utils import try_import
with try_import("HAS_TLX"):
    import triton.language.extra.tlx as tlx
    from .tlx_attn_ws_pipelined_pingpong_hopper import (
        attention as tlx_attn_ws_pipelined_pingpong_hopper,
    )
```

PR #326, opened 2025-08-07
Improve jagged_mean baseline

yf225:
`torch._segment_reduce`-based jagged_mean baseline is a much better baseline as it doesn't make other implementations look exponentially better as we scale the input size. This is important for more fair benchmarking such as for Helion project.


FindHao:
> `torch._segment_reduce`-based jagged_mean baseline is a much better baseline as it doesn't make other implementations look exponentially better as we scale the input size. This is important for more fair benchmarking such as for Helion project.

sorry, I don't quite get it. can you paste the results before and after you change the baseline?

PR #340, opened 2025-08-20
Add Support for TLX matmul

njriasan:
Summary: Adds a TLX matmul to TritonBench and guards it behind TLX support.

Differential Revision: D79178034

### DIFF
--- a/tritonbench/operators/gemm/operator.py
+++ b/tritonbench/operators/gemm/operator.py
@@ -16,6 +16,16 @@ from tritonbench.operators.gemm.warp_spec_persistent_matmul import (
     blackwell_matmul_tma,
     blackwell_matmul_tma_persistent,
 )
+from tritonbench.utils.triton_utils import has_tlx
+
+if has_tlx():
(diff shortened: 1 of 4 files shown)


njriasan:
@adamomainz The remaining linting issue is in another operator. To avoid having the scope of this PR explode I can do linting in another followup.

tritonbench/operators/gemm/operator.py
```diff
@@ -445,6 +455,13 @@ def triton_blackwell_descriptor_matmul(self, a, b, bias) -> Callable:
                     a, b, warp_specialize=False
                 )
 
+        @register_benchmark(enabled=False)
```
adamomainz, line 458:
```suggestion
        @register_benchmark(enabled=has_tlx())
```

might be better? do we want it auto enabled if we have tlx?

njriasan: My understanding was this can trigger CI at times so I always default to having it be False. Since our internal version doesn't fully work with TLX I will keep this disabled for now.

PR #341, opened 2025-08-20
Add amax as default per-row scaling factor for fp8_gemm benchmark

jananisriram:
Summary: Add `amax` (absolute maximum) as the default scaling factor for per-row scaling for fp8 GEMMs, as is used in practice.

Reviewed By: NikhilAPatel

Differential Revision: D80590746


xuzhao9:
The CI is broken for the fp8_gemm operator, can you help take a look? @jananisriram

PR #345, opened 2025-08-20
[RoPE] update input settings

BoyuanFeng:


xuzhao9:
The fp8_gemm CI failure should be unrelated. It seems to be inductor issue.

PR #356, opened 2025-08-25
Make the result consistent with FAv4 benchmark

xuzhao9:
Summary:
This Diff is the second attempt to close the gap between Tritonbench and FAv4 benchmark. It turns out that there are two reasons:

1. In FAv2 benchmark, it actually sets head_dim_kv = 2 where headdim = 16. In Tritonbench, it sets headdim = headdim_kv = 16.

2. Tritonbench uses rep = 100 by default where FAv2 benchmark uses rep = 10 for triton.testing.do_bench. I guess Tritonbench runs too may repeats that the GPU is power-throttled. Using --iter 10 will make the results consistent.

Differential Revision: D80984604


jduprat (approved):
This was tested and verified on a machine that showed significant discrepancy between TritonBench and FA4 benchmarking.  After this change we are within 0.2% tflops, which is plenty close enough.

PR #363, opened 2025-08-27
Enable sleeping between trials

njriasan:
Summary:
Adds command line support for a `--sleep` argument that can be used to inject a sleep in between trials. This is potentially useful on Blackwell where due to the floating clock measurements sometimes favor a earlier running kernel.

While most of this should be covered by just specifying running for at least 2 seconds, this can help ensure a more consistent run.

Reviewed By: xuzhao9, FindHao

Differential Revision: D81152631


FindHao (approved):
stamped for oss

PR #367, opened 2025-08-28
[ci] Add docker cleanup CI

xuzhao9:
We only keep the most recent 200 dockers.
Remove dockers that are not tagged.


FindHao (approved):
LGTM

PR #368, opened 2025-08-28
[gdpa] Add pingpong around tanh

manman-ren:
Summary: adjust registers, remove barrier_wait for o0/o1 inside the loop

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/tritonbench/operators/gdpa/gdpa_blackwell_tlx.py
+++ b/tritonbench/operators/gdpa/gdpa_blackwell_tlx.py
@@ -51,7 +51,7 @@ def get_cuda_autotune_config():
         for bkv in [3]
         for bqk in [1]  # in tmem
         for bo in [1]  # in tmem
-        for SUBTILE in [True] # doesn't support False
+        for SUBTILE in [True]  # doesn't support False
         for pp in [True, False]
         for ar in [192, 232]
     ]


tritonbench/operators/gdpa/gdpa_blackwell_tlx.py
```diff
@@ -505,32 +521,47 @@ def gdpa_kernel_tma_ws_blackwell(
...
+                        square = _mul_f32x2(qk0, qk0)
+                        inner = _fma_f32x2(c1, square, c0)
+                        inner0 = _mul_f32x2(inner, qk0)
+                        square = _mul_f32x2(qk1, qk1)
+                        inner = _fma_f32x2(c1, square, c0)
+                        inner1 = _mul_f32x2(inner, qk1)
+
+                        tlx.named_barrier_wait(10, 128)
+                        # p0 = fast_gelu(qk0)
+                        p0 = _fma_f32x2(qk0, tanh_approx_fp32(inner0), qk0)
```
manman-ren, line 552: I try to outline this section into a helper function, but hit
  File "/home/mren/MetaMain/triton/python/triton/language/extra/tlx/mem_ops.py", line 332, in local_reinterpret
    reinterpreted_value_handle = _builder.create_memdesc_reinterpret(src.handle, dtype.to_ir(_builder), shape)
                                                                                 ^^^^^^^^^^^
AttributeError: 'constexpr' object has no attribute 'to_ir'

Not sure if I need to mark dtype as const? @pchen7e2

tritonbench/operators/gdpa/gdpa_blackwell_tlx.py
```diff
@@ -375,7 +375,7 @@ def gdpa_kernel_tma_ws_blackwell(
 
     with tlx.async_tasks():
         # activation calculation
-        with tlx.async_task("default", registers=192):
+        with tlx.async_task("default", registers=232):
```
manman-ren: Another question is in terms of cleaning up the implementation:
can we use duplicate here for the default partition and another partition? @htyu

htyu: > Another question is in terms of cleaning up the implementation: can we use duplicate here for the default partition and another partition? @htyu

We may not, as the default region is unique. Can we use a non-default region for activations?

tritonbench/operators/gdpa/gdpa_blackwell_tlx.py
htyu: So `p0 = fast_gelu(qk0)` doesn't work with pingpong?

manman-ren: Only want to pingpong on the SFU/tanh.

htyu: Outlining the tlx.named_barrier_wait op causes the above compilation error?

PR #372, opened 2025-08-28
Refactor imports to conditionally include TLX based on availability

FindHao:
- Added a utility check for TLX support before importing `triton.language.extra.tlx` in both `gdpa_blackwell_tlx.py` and `tlx_matmul.py`.
fix internal test: test_gpu_tritonbench_gdpa


FindHao:
closed because of duplication

PR #375, opened 2025-08-28
[FA][Blackwell] Add triton version with autoWS

manman-ren:
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/tritonbench/operators/blackwell_attentions/operator.py
+++ b/tritonbench/operators/blackwell_attentions/operator.py
@@ -19,9 +19,15 @@ from torch.nn.functional import scaled_dot_product_attention as sdpa
 
 from tritonbench.kernels.attention_utils import SUPPORT_GLUON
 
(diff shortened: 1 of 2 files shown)


tritonbench/operators/blackwell_attentions/operator.py
```diff
@@ -19,6 +19,10 @@
 
 from tritonbench.kernels.attention_utils import SUPPORT_GLUON
 
+from tritonbench.kernels.blackwell_triton_fused_attention import (
+    attention_opt as blackwell_triton_tutorial_FA2_opt,
```
njriasan: I think this can break on import in Triton if anything doesn't work on all versions we support. Can we wrap this in a try catch with a comment that this may require the latest OSS Triton?

manman-ren: Good catch!

PR #380, opened 2025-09-01
Add missing torch.compile impl / improve compile config for several kernels

yf225:


tritonbench/operators/jagged_mean/operator.py
```diff
@@ -138,6 +147,15 @@ def torch_jagged_mean_torch_nanmean(
...
     @register_benchmark()
+    def torch_compile_jagged_mean_torch_nanmean(
+        self, x: torch.Tensor, B: int, M: int, seqlen: int, sparsity: float
+    ):
+        return torch.compile(
+            self.torch_jagged_mean_torch_nanmean(x, B, M, seqlen, sparsity),
+            mode="max-autotune-no-cudagraphs",
+        )
+
+    @register_benchmark(baseline=True)
```
yf225, line 158: This is generally the best eager implementation, so should be used as the eager baseline.

PR #384, opened 2025-09-02
[WIP][FA][Blackwell] Implementation with explicit data partitioning

manman-ren:
Currently hits "out of resource" as it needs buffer reuse to work


tritonbench/kernels/blackwell_triton_fused_attention_dp.py
```diff
@@ -0,0 +1,397 @@
...
+        # Use device_descriptor for Hopper + warpspec.
+        if supports_host_descriptor() and not (is_hopper() and warp_specialize):
+            # Note that on Hopper we cannot perform a FP8 dot with a non-transposed second tensor
+            y_dim = q.shape[0] * q.shape[1] * q.shape[2]
+
+            dummy_block = [1, 1]
+            desc_q = TensorDescriptor(
+                q,
+                shape=[y_dim, HEAD_DIM_K],
+                strides=[HEAD_DIM_K, 1],
```
njriasan, line 418: This assumes that q, k, and v are all contiguous and not transposed. Can we add an assert to enforce this requirement?

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
```diff
@@ -0,0 +1,397 @@
...
+                block_shape=dummy_block,
+            )
+            desc_o = TensorDescriptor(
+                o,
+                shape=[y_dim, HEAD_DIM_K],
+                strides=[HEAD_DIM_K, 1],
+                block_shape=dummy_block,
+            )
+        else:
```
njriasan, line 447: Just error on the else case? The kernel requires TMA

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
```diff
@@ -0,0 +1,397 @@
...
+              HEAD_DIM: tl.constexpr,  #
+              BLOCK_M: tl.constexpr,  #
+              BLOCK_N: tl.constexpr,  #
+              FP8_OUTPUT: tl.constexpr,  #
+              STAGE: tl.constexpr,  #
+              warp_specialize: tl.constexpr,  #
+              OUTER_LOOP: tl.constexpr,
+              dtype: tl.constexpr,
+              ):
+    n_tile_num = tl.cdiv(N_CTX, BLOCK_M)
```
njriasan, line 352: Is this correct? This stands out to me as a possible typo, possibly just in variable name.

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
```diff
@@ -0,0 +1,397 @@
...
+    num_progs = tl.num_programs(0)
+    total_tiles = n_tile_num * Z * H
+
+    tiles_per_sm = total_tiles // num_progs
+    if prog_id < total_tiles % num_progs:
+        tiles_per_sm += 1
+
+    tile_idx = prog_id
+    # inner loop warpspec vs. outer loop warpspec
+    for _ in tl.range(0, tiles_per_sm, warp_specialize=warp_specialize and OUTER_LOOP):
```
njriasan, line 363: For my understanding why don't we need `FLATTEN=True` here? Is it not viable with FA because its too complicated and we actually need more complex AutoWS in the compiler?

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
```diff
@@ -0,0 +1,397 @@
...
+    off_h = off_hz % H
+
+    offset_y = off_z + off_h * N_CTX
+    qo_offset_y = offset_y + start_m * BLOCK_M
+    # initialize offsets
+    offs_m0 = start_m * BLOCK_M + tl.arange(0, BLOCK_M//2)
+    offs_m1 = start_m * BLOCK_M + tl.arange(BLOCK_M//2, BLOCK_M)
+    offs_n = tl.arange(0, BLOCK_N)
+
+    m_i0 = tl.zeros([BLOCK_M//2], dtype=tl.float32) - float("inf")
```
njriasan: To clarify this is the explicit data partitioning to enable subtiling + ping pong?

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
```diff
@@ -0,0 +1,397 @@
...
+    # -- compute correction factor
+    alpha = tl.math.exp2(m_i - m_ij)
+    l_ij = tl.sum(p, 1)
+
+    # -- update output accumulator --
+    BM: tl.constexpr = acc.shape[0]
+    BN: tl.constexpr = acc.shape[1]
+
+    acc0, acc1 = acc.reshape([BM, 2, BN//2]).permute(0, 2, 1).split()
+    acc0 = acc0 * alpha[:, None]
```
njriasan, line 79: This is another form of partitioning/subtiling?

manman-ren: Yes, this is subtiling for correction.
FA tutorial has this too: https://github.com/triton-lang/triton/blob/main/python/tutorials/06-fused-attention.py#L88

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
manman-ren: This is to set up the outputs for data partitioning, we will call _attn_fwd_inner_oss_dp twice, each one working on one half of the full block size.
We will have one data partition working on q0, the other working on q1, where q0 + q1 is the original q.

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
manman-ren: FLATTEN only works on very simple case such as GEMM. For FA, FLATTEN doesn't work and we will need to handle nested control flow. In OSS, NVidia is driving the work.

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
manman-ren: This is to support on-device TMA. I will need to call _maybe_make_tensor_desc.

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
manman-ren: Yeah should rename n_tile_num to num_pid_m.

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
njriasan: Thanks for clarifying/confirming my understanding.

PR #391, opened 2025-09-03
Add `--cudagraph` support to `--latency-measure-mode=profiler`

yf225:
The cudagraph path will also clear L2 cache, which is important for getting accurate "cold-start" timing.


yf225:
> does inductor benchmarker support cudagraph?

I believe it doesn't at the moment, but since profiler with cudagraph is quite good and gives stable readings, maybe we can direct users to use it instead.

PR #392, opened 2025-09-04
Add Torch Benchmark

NikhilAPatel:
Summary:
Adds a benchmark for the new `torch._grouped_gemm` operator. This operator also requires all `B` tensors to be the same. To keep apples-to-apples comparisons, we’ll apply that constraint across the existing benchmarks as well. The ATen operator here is the measurement baseline, so our benchmarks should conform to its input contract

This restriction is lifted when using `torch._grouped_gemm` with 3D arguments. So in the future, we may want to create a separate Tritonbench operator for this.

Note: The Triton benchmark's accuracy was 0 before this diff and remains 0. This will be addressed in a followup.

Differential Revision: D81107417


xuzhao9 (approved):
LGTM!

PR #400, opened 2025-09-05
Add fp8 triton_mm to TritonBench

RandySheriff:
Summary: As titled ~ this is to tune fp8 triton_mm, which is a choice of torch._scaled_mm during lowering.

Reviewed By: beginner137

Differential Revision: D81554169


njriasan:
@RandySheriff  Is this different than https://github.com/meta-pytorch/tritonbench/tree/main/tritonbench/operators/fp8_gemm? Would it be possible to consolidate these? 

If this is rowwise or blockwise scaling then we could add it to those operators.

tritonbench/operators/fp8_gemm_rowwise/operator.py
```diff
@@ -198,6 +199,17 @@ def _triton(self, xq, wq, x_scale, w_scale, bias) -> Callable:
             use_warp_specialization=self.warp_specialization,
         )
 
+    @register_benchmark(enabled=HAS_TRITON)
```
njriasan, line 203: Are there any restrictions on where this can run e.g. H100 vs MI300? If not this looks good.

njriasan (approved):
LGTM! See my one comment but otherwise looks good.

PR #401, opened 2025-09-05
Profiler mode: Add ProfilerActivity.CPU to tracked activities

yf225:
When using profiler for latency measurement, I've been seeing cases where no CUDA kernel events were found in the profiler trace. I am still trying to root-cause this, but want to make the change in the PR as a defensive fix in case we are hitting some corner case when CPU activity not being recorded is causing a CUDA kernel event recording problem.

### DIFF
--- a/tritonbench/components/do_bench/run.py
+++ b/tritonbench/components/do_bench/run.py
@@ -227,7 +227,10 @@ def _do_bench_profiler(
     # Benchmark phase - collect kernel times for each iteration
     all_kernel_times = []
     profiler_config = {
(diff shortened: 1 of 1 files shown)


xuzhao9 (approved):

xuzhao9:
Can I get more details? What are the operators that are missing CUDA events?

yf225:
> Can I get more details? What are the operators that are missing CUDA events?

It's also very hard to reproduce - only happens after a few hours of running Helion benchmark, and only the last input shape have this problem. I am still trying to figure out why it happens and will share more if I figured it out :/

PR #405, opened 2025-09-08
Dump IRs of all Triton operators

xuzhao9:
Summary: Add `dump_ir`, a user benchmark that runs all custom Triton operators and dump their IRs to a directory. It will only keep the optimal IR after the autotuning.

Differential Revision: D81839643


benchmarks/dump_ir/run.py
```diff
@@ -0,0 +1,54 @@
...
+METADATA_DIR = parutil.get_file_path("tritonbench/metadata") if is_fbcode() \
+    else Path(__file__).parent.parent.parent.joinpath("tritonbench/metadata")
+
+OSS_CUSTOM_TRITON_YAML = os.path.join(METADATA_DIR, "oss_triton_operators.yaml")
+INTERNAL_CUSTOM_TRITON_YAML = os.path.join(METADATA_DIR, "fb/internal_triton_operators.yaml")
+
+
+def get_parser():
+    parser = argparse.ArgumentParser(description="Dump Triton IRs")
+    parser.add_argument("--scope", type=str, default="internal", choices=["oss", "internal"], help="Scope of the running operators.")
```
adamomainz: can this just be figured out by checking `is_fbcode` or something? we cant get the oss version internally right or are you thinking trunk?

adamomainz (approved):
left 1 nit

benchmarks/dump_ir/run.py
xuzhao9: Good point, let's use `is_fbcode()`  to determine the scope

PR #409, opened 2025-09-09
Strengthen Triton Benchmark

NikhilAPatel:
Summary: Updated configs to use the correct number of SMs, modify kernel to use the specified dtype, and allocate `c` with `torch.zeros` rather than `torch.empty` to mitigate the possibility of false positive accuracy checks.

Differential Revision: D81107418


jananisriram (approved):
lgtm!

PR #386, opened 2025-09-03
Add `--latency-measure-mode=profiler` support

yf225:
Having a latency measurement method that only measures GPU kernel runtime is important. I believe existing options such as Triton `do_bench` or inductor benchmarker still includes CPU overhead in the measurement, and `--cudagraph` doesn't work for all kernels. Using profiler to measure GPU kernel runtime is a good middle ground that gives us relatively reliable timing measurement.

Triton `do_bench` (includes CPU overhead):
```
      (M, N)    torch_layer_norm-latency    triton_layer_norm-latency    torch_compile_layer_norm-latency    liger_layer_norm-latency    helion_layer_norm_fwd-latency
------------  --------------------------  ---------------------------  ----------------------------------  --------------------------  -------------------------------
(4096, 1024)           0.023936 (±2.27%)            0.021888 (±6.87%)                   0.020288 (±7.73%)           0.020960 (±7.02%)                0.020064 (±4.63%)
```

`--cudagraph` (i.e. ground truth GPU kernel execution time):
```
      (M, N)    torch_layer_norm-latency    triton_layer_norm-latency    torch_compile_layer_norm-latency    liger_layer_norm-latency    helion_layer_norm_fwd-latency
------------  --------------------------  ---------------------------  ----------------------------------  --------------------------  -------------------------------
(4096, 1024)           0.014104 (±0.13%)            0.008505 (±0.12%)                   0.007519 (±0.08%)           0.006929 (±0.09%)                0.006659 (±0.06%)
```

Profiler (i.e. this PR):
```
      (M, N)    torch_layer_norm-latency    triton_layer_norm-latency    torch_compile_layer_norm-latency    liger_layer_norm-latency    quack_layer_norm-latency
------------  --------------------------  ---------------------------  ----------------------------------  --------------------------  --------------------------
(4096, 1024)           0.014304 (±1.36%)            0.008640 (±2.97%)                   0.007584 (±1.67%)           0.007168 (±3.57%)           0.008192 (±3.13%)
```

As seen above, the profiler approach is much closer to cudagraph (i.e. ground truth) wrt. measured GPU kernel execution time.

PR #411, opened 2025-09-10
[do_bench][easy] warmup cudagraph mode in do_bench_profiler

BoyuanFeng:
Prior to this PR, cudagraph mode does not warmup. 

However, both default mode and cudagraph mode need warmup. For a compiled function, CUDAGraph mode may record all autotune kernels if it is not warm up properly.

### DIFF
--- a/tritonbench/components/do_bench/run.py
+++ b/tritonbench/components/do_bench/run.py
@@ -193,7 +193,7 @@ def _do_bench_profiler(
 
     # Calculate number of iterations based on target rep time
     if estimate_ms == 0:
-        n_repeat = 100  # Default if function is very fast
+        n_repeat = 1000  # Default if function is very fast
     else:
         n_repeat = max(1, int(rep / estimate_ms))
 
@@ -205,6 +205,14 @@ def _do_bench_profiler(
         cache.zero_()
         fn()
 
+    # Warmup
+    n_warmup = max(1, int(warmup / estimate_ms)) if estimate_ms > 0 else 25
+
+    torch.cuda.synchronize()
+    for _ in range(n_warmup):
+        run_iteration()
+    torch.cuda.synchronize()
+
     if use_cudagraph:
         # Create CUDA graph
         g = torch.cuda.CUDAGraph()
@@ -212,17 +220,8 @@ def _do_bench_profiler(
             for _ in range(n_repeat):
                 run_iteration()
         torch.cuda.synchronize()
-    else:
-        # Regular mode warmup
-        n_warmup = max(1, int(warmup / estimate_ms)) if estimate_ms > 0 else 25
 
-        torch.cuda.synchronize()
-        for _ in range(n_warmup):
-            run_iteration()
-        torch.cuda.synchronize()
-
-    n_profiler_runs = 5
-    iterations_per_profiler_run = n_repeat
+    n_profiler_runs = 10
 
     # Benchmark phase - collect kernel times for each iteration
     all_kernel_times = []
@@ -243,7 +242,7 @@ def _do_bench_profiler(
                 g.replay()
             else:
                 # Execute multiple iterations for regular mode
-                for _ in range(iterations_per_profiler_run):
+                for _ in range(n_repeat):
                     run_iteration()
             torch.cuda.synchronize()
 
@@ -301,7 +300,7 @@ def _do_bench_profiler(
         # Convert to milliseconds and normalize by iterations
         total_kernel_time_ms = (
             total_kernel_time_us / 1000.0
-        ) / iterations_per_profiler_run
+        ) / n_repeat
         all_kernel_times.append(total_kernel_time_ms)
 
     times = torch.tensor(all_kernel_times, dtype=torch.float)


tritonbench/components/do_bench/run.py
```diff
@@ -205,24 +205,23 @@ def run_iteration():
         cache.zero_()
         fn()
 
+    # Warmup
+    n_warmup = max(1, int(warmup / estimate_ms)) if estimate_ms > 0 else 25
+
+    torch.cuda.synchronize()
+    for _ in range(n_warmup):
+        run_iteration()
+    torch.cuda.synchronize()
```
yf225: Wonder does `estimate_ms = benchmarker.benchmark_gpu(fn, estimation_iters=5, benchmark_iters=10)` above already serve the warmup purpose?

Or if it's not explicit enough, should we move this explicit warmup to be before `estimate_ms`?

BoyuanFeng: yes moved it to the top

tritonbench/components/do_bench/run.py
```diff
@@ -193,36 +201,28 @@ def _do_bench_profiler(
...
 
     # Helper function to execute one iteration
-    def run_iteration():
+    def run_iteration(should_clear_cache: bool):
         if grad_to_none is not None:
             for x in grad_to_none:
                 x.grad = None
-        cache.zero_()
+        if should_clear_cache:
+            cache.zero_()
```
BoyuanFeng: when measuring with cudagraph, we should not clear cache, since it adds extra memory access time.
This matches the behavior of triton.testing.do_bench_cudagraph
https://github.com/triton-lang/triton/blob/f90255886173b873dfb8b5bbb9a3f67951954660/python/triton/testing.py#L106-L111

xuzhao9: I thought we should still clear L2 cache when running with cudagraph. This is an issue with `do_bench_cudagraph` cc @nmacchioni

BoyuanFeng: we use tensor.zero_() to clear L2 cache. if we do that in cudagraph, the graph would contain both 1) the tensor.zero() AND 2) the fn to benchmark. Would this lead to wrong result?

By contract, in [do_bench](https://github.com/triton-lang/triton/blob/f90255886173b873dfb8b5bbb9a3f67951954660/python/triton/testing.py#L174C1-L186C30), we do clear_cache outside measurement which does NOT include the clearing L2 cache time.

```
    for i in range(n_repeat):
        # we don't want `fn` to accumulate gradient values
        # if it contains a backward pass. So we clear the
        # provided gradients
        if grad_to_none is not None:
            for x in grad_to_none:
                x.grad = None
        # we clear the L2 cache before each run
        runtime.driver.active.clear_cache(cache)
        # record time of `fn`
        start_event[i].record()
        fn()
        end_event[i].record()
```

xuzhao9: L2 cache clear is essential to more stable and accurate latency measurements.
We will explicitly exclude the latency of the `CACHE_CLEAR_KERNEL` from the kernel latencies:  https://github.com/BoyuanFeng/tritonbench/blob/a0c75ad868f9d03f1cf6107da139e500cfe0b60b/tritonbench/components/do_bench/run.py#L259

BoyuanFeng: this is for profiler-based benchmark but not for cudagraph-based benchmark?

xuzhao9: We provide cudagraph option for profiler-based benchmark, e.g., `--latency-measure-mode=profiler --cudagraph`: https://github.com/meta-pytorch/tritonbench/pull/386

yf225: echoing Xu, the `and evt.name != CACHE_CLEAR_KERNEL` below should exclude `tensor.zero_()` from the GPU trace, so I believe we should run `tensor.zero_()` even with cudagraph-based benchmark. Otherwise `--profiler --cudagraph` vs. `--profiler`-only will give different latency results.

BoyuanFeng: agree. `CACHE_CLEAR_KERNEL` would still appear in cuda trace when cudagraph is used. thanks for clarification!

tritonbench/components/do_bench/run.py
```diff
@@ -243,13 +242,28 @@ def run_iteration():
...
+        num_cache_clear_kernels = len(
+            [
+                evt
+                for evt in prof.events()
+                if evt.device_type == torch.autograd.DeviceType.CUDA
+                and evt.name == CACHE_CLEAR_KERNEL
+            ]
+        )
+        assert (
+            num_cache_clear_kernels == n_repeat
```
BoyuanFeng, line 264: assert to make sure L2 cache clear dispatch to hard-coded kernel name

xuzhao9: This makes sense to me, thanks!

tritonbench/components/do_bench/run.py
```diff
@@ -185,6 +185,14 @@ def _do_bench_profiler(
     Returns:
```
xuzhao9, line 185: After this PR, the parameter `warmup` will not be used by this function. Can you help remove it?

BoyuanFeng: let's keep it to match the API of `_do_bench_inductor` and `triton.testing.do_bench`

oulgen:
@BoyuanFeng this PR breaks helion: https://github.com/pytorch/helion/actions/runs/17808758986/job/50627008406#step:10:81

PR #414, opened 2025-09-11
Improve BWD / FWD_BWD accuracy checks; fix layer_norm bwd check

yf225:
Test: `python run.py --op layer_norm --bwd --num-inputs 1 --metrics accuracy`

### DIFF
--- a/tritonbench/operators/layer_norm/operator.py
+++ b/tritonbench/operators/layer_norm/operator.py
@@ -82,9 +82,37 @@ class Operator(BenchmarkOperator):
         return lambda: LigerLayerNormFunction.apply(x, weight, bias, eps)
 
     def get_bwd_fn(self, fwd_fn: Callable) -> Callable:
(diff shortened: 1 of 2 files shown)


tritonbench/operators/layer_norm/operator.py
```diff
@@ -82,9 +82,37 @@ def liger_layer_norm(self, *args):
         return lambda: LigerLayerNormFunction.apply(x, weight, bias, eps)
 
     def get_bwd_fn(self, fwd_fn: Callable) -> Callable:
+        from torch.utils._pytree import tree_map
+
+        # Run forward once to get output
         y = fwd_fn()
         dy = 0.1 * torch.randn_like(y)
-        return lambda: y.backward(dy, retain_graph=True)
```
yf225, line 87: I believe usually to check backward accuracy, we need to check the gradients of the forward input tensors. But the previous design doesn't allow us to do it.

PR #420, opened 2025-09-15
[power] Add initial power chart support

xuzhao9:
Add initial power chart support

```
python run.py --op softmax --power-chart $HOME/power-chart
```

temp chart

image

power chart

image

freq chart 

image


tritonbench/components/power/chart.py
```diff
@@ -0,0 +1,134 @@
...
+    POWER_OUTPUT_DIR = os.path.join(output_dir, benchmark_name)
+    if not os.path.exists(POWER_OUTPUT_DIR):
+        os.mkdir(POWER_OUTPUT_DIR)
+    stdout_file_path = os.path.join(POWER_OUTPUT_DIR, QUERY_STDOUT_FILE)
+    stderr_file_path = os.path.join(POWER_OUTPUT_DIR, QUERY_STDERR_FILE)
+    # Run the command
+    query_cmd = QUERY_COMMAND.format(
+        QUERY_FREQUENCY=QUERY_FREQUENCY, QUERY_DEVICE=_get_cuda_device_id()
+    ).split(" ")
+    with open(stdout_file_path, "w") as stdout_file, open(
```
FindHao, line 112: nit for future improvement:
it's better to read the output directly from stdout/stderror. the log file dumps can be async and for backup.

FindHao (approved):
LGTM!

tritonbench/components/power/chart.py
xuzhao9: We need to collect more data, e.g., whether there will be delay between the log and benchmark timestamp.

PR #429, opened 2025-09-16
[ci] Add back pt2_cutlass_matmul

xuzhao9:
`pt2_cutlass_matmul` has resumed its sanity, and we can add it back to CI testing.


FindHao (approved):
LGTM

PR #430, opened 2025-09-16
Improve FWD/BWD accuracy checks for rms norm

mengluy0125:
Summary: Titled

Differential Revision: D82499895


xuzhao9:
We should really promote this bwd accuracy checking to the base class, I will do that in a follow-up PR.

PR #440, opened 2025-09-17
Add quack layernorm to compare againt helion layernorm

mengluy0125:
Summary: We add quack layernorm to torchBench to compare against Helion layernorm to collect perf signals.

Differential Revision: D82683838

### DIFF
--- a/tritonbench/operators/layer_norm/operator.py
+++ b/tritonbench/operators/layer_norm/operator.py
@@ -41,6 +41,12 @@ except ModuleNotFoundError:
     LigerLayerNormFunction = None
     HAS_LIGER_KERNEL = False
 
+try:
+    from quack.quack_layernorm import layernorm as quack_layernorm
+    HAS_QUACK_KERNEL = True
+except ModuleNotFoundError:
+    HAS_QUACK_KERNEL = False
+
 
 class Operator(BenchmarkOperator):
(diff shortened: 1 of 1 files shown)


tritonbench/operators/layer_norm/operator.py
```diff
@@ -81,6 +86,12 @@ def liger_layer_norm(self, *args):
         (x, w_shape, weight, bias, eps) = args
         return lambda: LigerLayerNormFunction.apply(x, weight, bias, eps)
 
+    @register_benchmark(enabled=QuackLayerNorm)
+    def quack_layernorm(self, *args) -> Callable:
+        (x, w_shape, weight, bias, eps) = args
+        quack_layernorm_op = QuackLayerNorm()
```
xuzhao9: Just curious why this has to be wrapped inside a `torch.nn.Module`?

xuzhao9 (approved):

tritonbench/operators/layer_norm/operator.py
mengluy0125: > Just curious why this has to be wrapped inside a `torch.nn.Module`?

I tried to keep a cleaner import structure. But you are right, removing the wrapper can be a better choice. Have updated the code.

PR #444, opened 2025-09-18
Use get_fp8_constants from fp8_utils.py instead of fbgemm_gpu

yf225:
Use get_fp8_constants from fp8_utils.py instead of fbgemm_gpu

### DIFF
--- a/tritonbench/utils/fp8_utils.py
+++ b/tritonbench/utils/fp8_utils.py
@@ -1,0 +1,51 @@
+"""FP8 utilities for tritonbench operators."""
+
+import functools
+import os
+from typing import Tuple
+
+import torch
+import triton.language as tl
+
+
+@functools.lru_cache
+def supports_float8_fnuz(throw_on_hip_incompatibility: bool = True) -> bool:
+    if torch.version.hip:
+        device_capability = torch.cuda.get_device_capability()
+
+        if device_capability < (9, 4):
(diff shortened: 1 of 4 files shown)


tritonbench/utils/fp8_utils.py
```diff
@@ -0,0 +1,51 @@
...
+    Helper function to get constant values for the current platform.
+
+    Returns:
+        pt_dtype (torch.dtype): The correct torch fp8 datatype.
+        tl_dtype (tl.dtype): The correct triton fp8 datatype.
+        max_fp8 (float): The maximum reprsentable value for the fp8 datatype.
+        eps (float): Minimum clip value to prevent divide by zero.
+    """
+    running_on_github: bool = os.getenv("GITHUB_ENV") is not None
+    if supports_float8_fnuz(throw_on_hip_incompatibility=(not running_on_github)):
```
xuzhao9, line 44: I am trying to understand, why github env needs to be handled separately?

yf225: It's a direct copy of https://github.com/pytorch/FBGEMM/blob/fd32631d837b41251311099c393af7d7be5cfbf5/fbgemm_gpu/experimental/gemm/triton_gemm/fp8_gemm.py#L35 - since we've been importing `get_fp8_constants` from there, I was worried that this env var is actually being relied on in OSS or internal CI tests

xuzhao9 (approved):

PR #452, opened 2025-09-18
[Fix] gather_gemv small fix

Sibylau:
- keep only one baseline=True for the Triton kernel
- fix the return type for result comparison

Thanks so much for suggested fix by @yf225


yf225 (approved):

tritonbench/operators/gather_gemv/operator.py
```diff
@@ -42,7 +42,7 @@ def __init__(
     def test_0(self, p1, p2, p3) -> Callable:
         return lambda: triton_test_0(p1, p2, p3)
 
-    @register_benchmark(baseline=True)
```
yf225, line 45: Sorry actually let's use `test_eager` as baseline

yf225 (changes requested):

PR #455, opened 2025-09-19
gather_gemv: fix eager impl to support large input; fix compile impl

yf225:


Sibylau:
Thank you @yf225 !! It solves the issue!

xuzhao9 (approved):

PR #459, opened 2025-09-20
Profiler mode: fix missing CUDA events

yf225:
Use profiler `trace_handler` to ensure that trace is ready before kernel runtime computation.

### DIFF
--- a/tritonbench/components/do_bench/run.py
+++ b/tritonbench/components/do_bench/run.py
@@ -236,17 +236,7 @@ def _do_bench_profiler(
         "with_stack": False,
     }
 
-    for _ in range(n_profiler_runs):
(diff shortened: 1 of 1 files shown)


xuzhao9 (approved):

xuzhao9:
Just curious, why would there be missing CUDA events before this PR? I thought after calling `torch.cuda.synchronize()`, all CUDA events should have been completed.

yf225:
> Just curious, why would there be missing CUDA events before this PR? I thought after calling `torch.cuda.synchronize()`, all CUDA events should have been completed.

Yeah I am really not sure why either.. It might have something to do with CUPTI buffer draining etc.

PR #462, opened 2025-09-22
[gdpa][proton] Add proton regions, also add an epilogue partition for tma_store

manman-ren:
ENABLE_PROTON=1 python third_party/tlx/tutorials/blackwell-gdpa.py
use regular barrier (not named barrier) for pingpong

### DIFF
--- a/tritonbench/operators/gdpa/gdpa_blackwell_tlx.py
+++ b/tritonbench/operators/gdpa/gdpa_blackwell_tlx.py
@@ -607,7 +607,9 @@ def gdpa_kernel_tma_ws_blackwell(
                             if NAME_BARRIER:
                                 tlx.named_barrier_wait(10, 128)
                             else:
-                                tlx.barrier_wait(producer_commit_pp[0], phase)  # consumer_wait
+                                tlx.barrier_wait(
+                                    producer_commit_pp[0], phase
+                                )  # consumer_wait
                         # p0 = fast_gelu(qk0)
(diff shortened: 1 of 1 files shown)


tritonbench/operators/gdpa/gdpa_blackwell_tlx.py
```diff
@@ -1366,6 +1512,10 @@ def grid_tma_persistent(META):
         IS_DENSE_KV=is_dense_kv,
         activation_enum_int=activation_enum_int,
         USE_ON_DEVICE_TMA=USE_ON_DEVICE_TMA,
+        NAME_BARRIER=False,
+        MERGE_EPI=False,
+        ENABLE_PROTON=enable_proton,
+        PROTON_TILE=10,
```
dshi7, line 1518: @fywkevin @manman-ren is it generally recommended to enable proton on single tile?

fywkevin: Single tile should be fine. In this case, we have very limited SMEM resource. So we restrict ourselves just dump a tile. Usually, it is enough to observe some interesting patterns.

dshi7: In terms of API usage, do you have plan to change the pair of `<pl.enter..., pl.exit...>` as a context manager, which might be more user friendly IMO?

PR #466, opened 2025-09-23
rms_norm: pass weight as input args, to reuse the same weight for all impls

yf225:
This is important for all rms_norm impls to share the same weights and pass the accuracy check.


oulgen (approved):

yf225:
Will merge this to unblock rms_norm backward testing.

Issue #468, opened 2025-09-24
[nsys_rep] NSYS will not work with nvtx


xuzhao9:
We want to use the following command to collect the nsys report file:

```
nsys profile -c nvtx --nvtx-capture tritonbench_range -t nvtx,osrt,cuda,cudnn,cublas -w true -f true -o nsys_rep.nsys-rep -- python run.py --op gemm --only aten_matmul --num-inputs 1 --input-id 0 --metrics _nsys_rep_in_task
```

However, even we set nvtx range (https://github.com/meta-pytorch/tritonbench/blob/main/tritonbench/components/ncu/__init__.py#L69), Nsys could not capture the kernels in the process, and it says "No reports were generated".

~~On Hopper, we can workaround this issue by setting capture mode to cudaProfilerApi, and we set both nvtx range and cudaproilerrange with `torch.cuda.cudart().cudaProfilerStart()` But on blackwell, neither cudaProfilerApi nor nvtx range will work, and we have to set the capture mode to "none" in order to generate the nsys report that contains the CUDA kernel:~~ 

Previously, we workaround this problem by setting capture mode to `cudaProfilerApi`. However, recently we found that this mode will not work for certain Meta internal kernel. And for this specific kernel, `cudaProfilerApi` won't work on either hopper or blackwell. Therefore, we have to change the capture model to `none` to capture the entire process.

Still, we want to understand why nsys won't work with nvtx.

Workaround in https://github.com/meta-pytorch/tritonbench/pull/469

Commands are tested using the latest nsight systems 2025.5.1
Note that the nsight compute (`ncu`) tool works fine with nvtx range, so there is no need to capture with `cudaProfilerApi`.

tcourtneynv:
Hi @xuzhao9 , if you are using `--capture-range` to capture an NVTX range (`-c nvtx`) then you should also specify the specific NVTX range to capture using `--nvtx-capture <range_name>`. I don't see that option on the first command line. Can you double check that you set `--nvtx-capture`?

I am not aware of any issues that would cause capture-range to behave differently between Hopper and Blackwell. Can you confirm you are using 2025.5.1 on both these systems? Are the NVTX ranges in Python or C/C++ code?  If Python, are you specifying the NVTX ranges using the NVTX support in pytorch, or with the NVTX Python package (https://pypi.org/project/nvtx/)? Does your code on both platforms have the same `cudaProfilerStart()` and `cudaProfilerStop()` calls?

[...]

PR #469, opened 2025-09-24
Set capture range to none

xuzhao9:
Summary: I am not sure why, but on blackwell, cudaProfilerApi won't work with nsys (nvtx does not work either), so we have to capture the entire process.

Differential Revision: D83152324


xuzhao9:
Described the problem in https://github.com/meta-pytorch/tritonbench/issues/468, pending NVIDIA to reproduce

PR #471, opened 2025-09-24
[docker] install cu130 toolchain

xuzhao9:
Upstream pytorch has CUDA 13.0 support


xuzhao9:
We have to hold the upgrade, because flash_attention does not support CUDA 13.0 yet.

Update 2026.04.07: Now the problem has been resolved.

PR #472, opened 2025-09-25
jagged_sum accuracy small fix

Sibylau:
Stacked PRs:
 * #484
 * __->__#472

--- --- ---

### DIFF
--- a/tritonbench/operators/jagged_sum/operator.py
+++ b/tritonbench/operators/jagged_sum/operator.py
@@ -218,10 +218,9 @@ class Operator(BenchmarkOperator):
                 batch_size, max_seq_len, _ = dense_0.shape
                 yield (nested_tensor, batch_size, 1, max_seq_len, 0.0)
 
-    @register_metric()
-    def accuracy(self, fn: Callable, baseline_fn: Callable) -> bool:
-        output = fn()["output"]
-        baseline_output = baseline_fn()["output"]
+    def _get_accuracy(self, fn: Callable, baseline_fn: Callable) -> bool:
+        output = fn()
+        baseline_output = baseline_fn()
         return torch.allclose(
             output, baseline_output, atol=ABSOLUTE_TOLERANCE, rtol=RELATIVE_TOLERANCE
         )
--- a/tritonbench/operators/jagged_sum/operator.py
+++ b/tritonbench/operators/jagged_sum/operator.py
@@ -102,8 +102,8 @@ class Operator(BenchmarkOperator):
         self, tb_args: argparse.Namespace, extra_args: Optional[List[str]] = None
     ):
         super().__init__(tb_args, extra_args)
-        self.sizes = (
-            list(range(2, 12, 4)) + list(range(11, 18, 3))
+        self.sizes = list(
+            range(2, 12, 4)
         )  # bias towards larger sizes, which are more representative of real-world shapes
 
         args = parse_op_args(self.extra_args)
--- a/tritonbench/utils/jagged_utils.py
+++ b/tritonbench/utils/jagged_utils.py
@@ -116,12 +116,12 @@ def generate_input_vals(B, M, max_seqlen, sparsity, sizes):
         B_vals.extend([B])
 
     if M is None:
-        M_vals.extend(get_dim_vals(sizes))
+        M_vals.extend([512])
     else:
         M_vals.extend([M])
 
     if max_seqlen is None:
-        seqlen_vals.extend((100, 500, 1000, 3000, 10_000, 18_000))
+        seqlen_vals.extend([512])
     else:
         seqlen_vals.extend([max_seqlen])
 


tritonbench/operators/jagged_sum/operator.py
```diff
@@ -218,10 +218,9 @@ def get_input_iter(self) -> Generator:
                 batch_size, max_seq_len, _ = dense_0.shape
                 yield (nested_tensor, batch_size, 1, max_seq_len, 0.0)
 
-    @register_metric()
-    def accuracy(self, fn: Callable, baseline_fn: Callable) -> bool:
-        output = fn()["output"]
-        baseline_output = baseline_fn()["output"]
+    def _get_accuracy(self, fn: Callable, baseline_fn: Callable) -> bool:
+        output = fn()
+        baseline_output = baseline_fn()
```
yf225, line 223: Actually is `def accuracy` without `@register_metric()` the right function to use? Which would match other operator classes as well

Sibylau: The original version with `@register_metric()` runs into errors. So I changed it to match the style in `jagged_mean`: https://github.com/meta-pytorch/tritonbench/blob/main/tritonbench/operators/jagged_mean/operator.py#L248 and it works

xuzhao9: Can you share the exact error? IMO we should fix it because `@register_metric` is the suggested user interface.

Sibylau: Yeah, the error is
```
Traceback (most recent call last):
  File "/home/jieeliu/workspace/tritonbench/tritonbench/utils/triton_op.py", line 981, in run
    y_vals: Dict[str, BenchmarkOperatorMetrics] = functools.reduce(
                                                  ^^^^^^^^^^^^^^^^^
  File "/home/jieeliu/workspace/tritonbench/tritonbench/utils/triton_op.py", line 966, in _reduce_benchmarks
    acc[bm_name] = self._do_bench(
                   ^^^^^^^^^^^^^^^
  File "/home/jieeliu/workspace/tritonbench/tritonbench/utils/triton_op.py", line 1466, in _do_bench
    self.accuracy(fn, self.baseline_fn) if self.baseline_fn else None
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jieeliu/workspace/tritonbench/tritonbench/utils/triton_op.py", line 640, in _inner
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jieeliu/workspace/tritonbench/tritonbench/operators/jagged_sum/operator.py", line 223, in accuracy
    output = fn()["output"]
             ~~~~^^^^^^^^^^
IndexError: too many indices for tensor of dimension 2
  x_val
```
and if I remove `["output"]`, I run into
```
Traceback (most recent call last):
  File "/home/jieeliu/workspace/tritonbench/tritonbench/utils/triton_op.py", line 981, in run
    y_vals: Dict[str, BenchmarkOperatorMetrics] = functools.reduce(
                                                  ^^^^^^^^^^^^^^^^^
  File "/home/jieeliu/workspace/tritonbench/tritonbench/utils/triton_op.py", line 966, in _reduce_benchmarks
    acc[bm_name] = self._do_bench(
                   ^^^^^^^^^^^^^^^
  File "/home/jieeliu/workspace/tritonbench/tritonbench/utils/triton_op.py", line 1466, in _do_bench
    self.accuracy(fn, self.baseline_fn) if self.baseline_fn else None
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jieeliu/workspace/tritonbench/tritonbench/utils/triton_op.py", line 640, in _inner
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jieeliu/workspace/tritonbench/tritonbench/operators/jagged_sum/operator.py", line 225, in accuracy
    return torch.allclose(
           ^^^^^^^^^^^^^^^
TypeError: allclose(): argument 'input' (position 1) must be Tensor, not NoneType
  x_val
```

Sibylau: btw, some of the generated tensor shapes for jagged kernels (`generate_input_vals` in jagged_utils.py) are too big to fit onto GPUs, and the test hangs on my end. I changed to smaller shapes locally but did not push.

yf225: Are you using H100? It might hang on Benchmark CI machines too so sgtm if we also reduce the input shape range (maybe in another PR and also get Xu's review)

Sibylau: It hangs on both my local H100 and B200... Sure I'll open another PR to reduce input shape range.

jananisriram: interesting that @register_metric accuracy is failing, since I also recall that to be the suggested interface. another operator [`vector_exp`](https://www.internalfb.com/code/fbsource/[256e3ee3c7f973e9b5e3ecbb00a23fc0ecaf8bcd]/fbcode/pytorch/tritonbench/tritonbench/operators/vector_exp/operator.py?lines=101) seems to also use this interface; might be worth changing that if we see the same error there? however, `_get_accuracy` seems to be the standard now; other operators also simply override the `def accuracy` function from `triton_op.py`

Issue #366, opened 2025-08-28
[BE][Help wanted] Automatically apply tags to backends


xuzhao9:
We want to provide API to deal with user queries like:

- all custom Triton kernels
- all PT2-Triton kernels
- all Triton kernels, custom and PT2
- all CUTLASS kernels
- all operators that has both Triton and CUTLASS
- ...

To do that, we need to apply tags to kernels, like "cutlass", "triton", "pt2-triton", "custom-triton", etc.

xuzhao9:
Addressed in https://github.com/meta-pytorch/tritonbench/pull/620

PR #474, opened 2025-09-25
More accurate kernel impl name prefix, for easier matching

yf225:
To make it easier for users to select only specific kernels to run.

### DIFF
--- a/tritonbench/operators/gather_gemv/operator.py
+++ b/tritonbench/operators/gather_gemv/operator.py
@@ -39,11 +39,11 @@ class Operator(BenchmarkOperator):
         super().__init__(tb_args, extra_args)
 
     @register_benchmark()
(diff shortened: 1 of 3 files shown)


oulgen:
@yf225 @xuzhao9 can we require that all tritonbench kernels have a particular naming scheme? i.e. helion kernels must be `helion_[kernel_name]`, torch compiles ones `torch_compile_[kernel_name]` etc?

xuzhao9:
@oulgen 

We want to address this with two features
1. We want these to be generated and maintained in metadata yaml like https://github.com/meta-pytorch/tritonbench/blob/main/tritonbench/metadata/oss_triton_operators.yaml

2. Additionally, user can attach tags like `@register_benchmark(tags=["cuda", ...])` to complement with the yaml file.
See https://github.com/meta-pytorch/tritonbench/issues/366

PR #477, opened 2025-09-25
cross_entropy: modify Liger triton kernel to be cudagraph compatible

yf225:
Previously, running cross_entropy Liger kernels with Triton do_bench_cudagraph fails with `CUDA error: operation failed due to a previous error during capture`. This PR removes the incompatibility from the Liger kernel to make it cudagraph compatible.

Performance of the Liger kernel stays the same:
On H100:
Before:
```
$ python run.py   --op cross_entropy --metrics speedup,accuracy               --latency-measure-mode profiler               --only liger               --only-match-mode prefix-with-baseline
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 6/6 [00:20<00:00,  3.35s/it]
        (B, T, V)    liger_cross_entropy_loss-speedup    liger_cross_entropy_loss-accuracy
-----------------  ----------------------------------  -----------------------------------
  (8, 2048, 4096)                            0.848115                                    1
  (8, 2048, 8192)                            1.12636                                     1
 (8, 2048, 16384)                            0.909232                                    1
 (8, 2048, 32768)                            1.48511                                     1
 (8, 2048, 65536)                            1.55754                                     1
(8, 2048, 131072)                            1.3279                                      1
          average                            1.20904                                     1
```

After:
```
$ python run.py   --op cross_entropy --metrics speedup,accuracy \
              --latency-measure-mode profiler \
              --only liger \
              --only-match-mode prefix-with-baseline
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 6/6 [00:23<00:00,  3.95s/it]
        (B, T, V)    liger_cross_entropy_loss-speedup    liger_cross_entropy_loss-accuracy
-----------------  ----------------------------------  -----------------------------------
  (8, 2048, 4096)                            0.84033                                     1
  (8, 2048, 8192)                            1.10772                                     1
 (8, 2048, 16384)                            0.903518                                    1
 (8, 2048, 32768)                            1.47731                                     1
 (8, 2048, 65536)                            1.55918                                     1
(8, 2048, 131072)                            1.32759                                     1
          average                            1.20261                                     1
```

### DIFF
--- a/tritonbench/operators/cross_entropy/liger_cross_entropy.py
+++ b/tritonbench/operators/cross_entropy/liger_cross_entropy.py
@@ -1,0 +1,542 @@
+import operator
+
+from typing import Optional
+
+import torch
+import torch.nn as nn
+import triton
+import triton.language as tl
+
+from liger_kernel.ops.utils import compare_version, element_mul_kernel, is_hip
+from liger_kernel.utils import infer_device
+
+if compare_version("triton", operator.ge, "3.0.0"):
+    try:
+        # typical import path with dispatch available
+        from triton.language.extra.libdevice import tanh
+    except ModuleNotFoundError:
+        # for working with NGC containers
+        from triton.language.extra.cuda.libdevice import tanh
+else:
+    from triton.language.math import tanh
+
+
+@triton.jit
+def liger_cross_entropy_kernel(
+    X_ptr,
+    X_stride,
+    Y_ptr,
+    Y_stride,
+    weight_ptr,
+    loss_ptr,
+    z_loss_ptr,
+    loss_stride,
+    n_cols,
+    n_non_ignore_ptr,
+    sum_non_ignore_weight_ptr,
+    weight_sum_ptr,
+    ignore_index,
+    lse_square_scale: tl.constexpr,
+    label_smoothing: tl.constexpr,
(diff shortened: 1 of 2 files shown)


xuzhao9 (approved):
It is good for now, but in longer term it will lose track with upstream since the upstream Liger kernel repo might change.

What exactly makes the upstream Liger kernel incompatible with CUDA Graph? Is it because of dynamic memory allocation?

yf225:
> It is good for now, but in longer term it will lose track with upstream since the upstream Liger kernel repo might change.
> 
> What exactly makes the upstream Liger kernel incompatible with CUDA Graph? Is it because of dynamic memory allocation?

These are the incompatibilities I found:
- Upstream still pulls GPU scalars back to Python during the forward path: n_non_ignore = target_mask.sum().item() forces a sync when the graph is being captured, which CUDA Graph prohibits (liger_kernel/ops/cross_entropy.py:288).
- The same host sync happens for the class-weight accumulators: sum_non_ignore_weight = ... .sum().item() and weight_sum = weight.sum().item() both materialize device values on the CPU during capture (liger_kernel/ops/cross_entropy.py:300, liger_kernel/
  ops/cross_entropy.py:301).

I'll follow up to file an issue / a PR to Liger repo.

PR #484, opened 2025-09-27
change jagged tensor generation to smaller shapes

Sibylau:
change jagged tensor generation to smaller shapes


tritonbench/utils/jagged_utils.py
```diff
@@ -116,17 +116,17 @@ def generate_input_vals(B, M, max_seqlen, sparsity, sizes):
...
 
     if max_seqlen is None:
-        seqlen_vals.extend((100, 500, 1000, 3000, 10_000, 18_000))
+        seqlen_vals.extend([512])
     else:
         seqlen_vals.extend([max_seqlen])
 
     if sparsity is None:
-        sparsity_vals.extend((0.1, 0.25, 0.5, 0.75, 0.9))
+        sparsity_vals.extend([0.5])
```
yf225: Wonder are all changes in this file required for making jagged kernel tests run without OOM on H100?

I think in general having more diversity in the input is still good and preferred, but if they OOM on H100 then we should reduce the range.

Sibylau: The sparsity does not lead to OOM. Thanks and corrected!

xuzhao9:
Hi @jananisriram, can you also help take a look at this PR, since you authored the input generator of the jagged operators?

jananisriram (approved):
looks good! out of curiosity, do you happen to know which function (if any) the `_get_accuracy` function here is overriding? curious to understand why `@register_metric` no longer works

xuzhao9:
@jananisriram Users hit an error at https://github.com/meta-pytorch/tritonbench/pull/472, I can take a look. Using `@register_metric` is recommended because `_get_accuracy()` is not user-facing.

xuzhao9 (approved):

PR #485, opened 2025-09-27
Skip benchmarks that are not enabled in --only mode

jananisriram:
Summary: Skip benchmarks that are not `enabled` in TritonBench when in `--only` mode unless `--force` is enabled.

Reviewed By: xuzhao9

Differential Revision: D83290749


xuzhao9 (approved):
LGTM!

PR #487, opened 2025-09-27
add non-persistent variant with data partitioning

manman-ren:
Summary: guard subtiling with tunable configure

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/tritonbench/kernels/blackwell_triton_fused_attention_dp.py
+++ b/tritonbench/kernels/blackwell_triton_fused_attention_dp.py
@@ -207,6 +207,7 @@ configs = [
         num_stages=s,
         num_warps=w,
         pre_hook=_host_descriptor_pre_hook,
+        # ir_override=f"/home/mren/OpenSource/tritonbench/override/_attn_fwd_persist.ttgir"
     )
     for BM in [256]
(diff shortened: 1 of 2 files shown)


tritonbench/kernels/blackwell_triton_fused_attention_dp.py
```diff
@@ -230,10 +262,11 @@ def _attn_fwd_tma_dp(
     STAGE: tl.constexpr,  #
     warp_specialize: tl.constexpr,  #
     dtype: tl.constexpr,
+    SUBTILING: tl.constexpr,
 ):
     tl.static_assert(BLOCK_N <= HEAD_DIM)
-    start_m = tl.program_id(0)
-    off_hz = tl.program_id(1)
+    start_m = pid  # tl.program_id(0)
+    # off_hz = tl.program_id(1)
```
njriasan, line 270: Is this change intentional?

manman-ren: Yes this is to fix existing issues. We passed in pid and off_hz, so we should use them. Otherwise, persistent will not work.

PR #488, opened 2025-09-29
Add persistent FA for blackwell

neildhar:
Modify the Blackwell FlashAttention implementation to support running a persistent kernel.


manman-ren (approved):
Thanks!

PR #489, opened 2025-09-29
Add benchmark post hook

yf225:
Add `benchmark_post_hook` which is called after each input benchmark completes. This is used by Helion benchmarker to print the problematic config that causes tritonbench accuracy check failure (Helion PR: https://github.com/pytorch/helion/pull/716).


xuzhao9:
Great! This can also be used for other purposes, e.g., for Mojo kernel accuracy checking we need to cast the output to fp32 in numpy format and compare it with torch.

PR #492, opened 2025-09-29
Print exception message when accuracy check fails

yf225:
This helps surface the root cause of the accuracy check error (e.g. numerical mismatch or # outputs mismatch etc.)

### DIFF
--- a/tritonbench/utils/triton_op.py
+++ b/tritonbench/utils/triton_op.py
@@ -1481,7 +1481,7 @@ class BenchmarkOperator(metaclass=PostInitProcessor):
                 )
             return True
         except Exception as e:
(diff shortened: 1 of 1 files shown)


tritonbench/utils/triton_op.py
```diff
@@ -1480,8 +1480,8 @@ def accuracy(self, fn: Callable, baseline_fn: Callable) -> bool:
                     atol=self.tb_args.atol,
                 )
             return True
-        except Exception:
-            # either the output tensor or the grad tensor does not match
+        except Exception as e:
+            print(f"Exception during accuracy check: {e}", file=sys.stderr)
```
xuzhao9: Can we use `logger.warning()` instead?

xuzhao9 (approved):

PR #494, opened 2025-09-29
[Grouped Gemm][CuTeDSL] Update default autotuning config set

NikhilAPatel:
Summary:
Update the set of default tuned CuTeDSL kernel configs to a performant set found by exhaustively autotuning on Qwen3 inference shapes.

The figure below shows the change in speedup over ATen gained from this change.
kde_comparison_ATen

Differential Revision: D83513302

### DIFF
--- a/tritonbench/operators/grouped_gemm/cutedsl/autotuner.py
+++ b/tritonbench/operators/grouped_gemm/cutedsl/autotuner.py
@@ -15,6 +15,8 @@ from functools import cached_property, partial
 from pathlib import Path
 from typing import Dict, List, Tuple
 
+import cutlass.utils as utils
+
 import torch
 
 import triton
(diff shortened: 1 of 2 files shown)


xuzhao9 (approved):
Amazing!

xuzhao9:
Can we also check-in the Qwen3 input shapes to https://github.com/meta-pytorch/tritonbench/tree/main/tritonbench/data/input_configs , if they are publicly available?

NikhilAPatel:
@xuzhao9 Great idea! They are technically public info, but they're not released by Qwen's company. Instead I manually logged them from running some inference benchmarks in vLLM. So let me know if you still think I should add them. Is there some script that can transform my list of shapes into the format of the jsons I see in this folder, or do I have to do it myself?

PR #495, opened 2025-09-30
Add TLX FA fwd kernel

htyu:
Summary: Adding TLX FA fwd kernel to tritonBench. The original kernel is in https://github.com/facebookexperimental/triton/blob/main/third_party/tlx/tutorials/blackwell-fa-ws-pipelined_test.py

Differential Revision: D83524653


xuzhao9 (approved):
Great work!

PR #502, opened 2025-10-01
[CI][AMD] Run the benchmark on MI350

xuzhao9:
Get the first set of MI350 benchmark data.


:wq
```diff
@@ -0,0 +1,136 @@
+name: linux-benchmark-mi350
```
FindHao: is the file name a typo?

xuzhao9: Yeah.. thanks for the catch!

xuzhao9:
Can I get another review, @FindHao ?

FindHao (approved):
Lgtm

PR #505, opened 2025-10-02
Load input shapes from highway_self_gating

xuzhao9:
Summary: Load input shapes from highway_self_gating traces.

Differential Revision: D83660588


FindHao (approved):
LGTM

PR #507, opened 2025-10-02
Move aten input loader to the new loader style

xuzhao9:
Summary:
We are refactoring the input loader component.

Now, each operator must implement its input loader script in the `input-loader.py` under its directory. This file must provide a function, `get_input_loader(tritonbench_op: Any, op: str, input: str)`, which will be plugged in as the new input generator.

The same interface holds for operator loaders like aten.

Reviewed By: FindHao

Differential Revision: D83754693


xuzhao9:
@FindHao , can I get a review?

PR #510, opened 2025-10-02
Print failing input id when error

yf225:
Making it easier to repro the error.


tritonbench/utils/triton_op.py
```diff
@@ -1116,6 +1116,11 @@ def _reduce_benchmarks(acc, bm_name: str):
                 "Caught exception, terminating early with partial results",
                 exc_info=True,
             )
+            if getattr(self, "_cur_input_id", None) is not None:
```
xuzhao9, line 1129: Maybe we also want to print the current benchmarking backend name?

PR #513, opened 2025-10-03
Do the "preprocessing" right for PyTorch compiled grouped GEMM

alexsamardzic:


alexsamardzic:
Currently, `torch_compile_grouped_gemm` and `preprocessed_pt2_triton_grouped_mm` are the same.

I think there is no point to benchmark `torch_compile_grouped_gemm`, as it takes into account both the auto-tuning and "preprocessing" arguments.   On the other hand, with the change in this PR, `preprocessed_pt2_triton_grouped_mm` is on par with `preprocessed_aten_grouped_mm`, which is expected.

(I believe the point about auto-tuning holds for `triton_grouped_gemm` too.)

xuzhao9:
cc @NikhilAPatel, can you help take a look?

xuzhao9:
Thanks!

PR #516, opened 2025-10-04
Do not hard fail when a benchmark fails

oulgen:
Do not hard fail when a benchmark fails


yf225:
wonder would removing `--exit-on-exception` CLI arg usage from Helion benchmark runner have the same effect?

oulgen:
> wonder would removing `--exit-on-exception` CLI arg usage from Helion benchmark runner have the same effect?

No because in my local testing it still hard fails

tritonbench/utils/triton_op.py
```diff
@@ -1091,14 +1091,20 @@ def _reduce_benchmarks(acc, bm_name: str):
...
+                            fn_name=bm_name,
+                            warmup=warmup,
+                            rep=rep,
+                            quantiles=quantiles,
+                            baseline=baseline,
+                        )
+                    except Exception:
+                        logger.exception(f"Name:{bm_name}\tInput Id:{input_id} failed")
+                        acc[bm_name] = BenchmarkOperatorMetrics(
+                            speedup=0, accuracy=False, extra_metrics=dict()
```
xuzhao9, line 1106: Can we log the `BenchmarkOperatorMetrics.error_msg` field? https://github.com/meta-pytorch/tritonbench/blob/161c9033ceee2609eadba1025e2385ad96a71841/tritonbench/utils/triton_op.py#L247

Similar to "CUDA OOM": https://github.com/meta-pytorch/tritonbench/blob/161c9033ceee2609eadba1025e2385ad96a71841/tritonbench/utils/triton_op.py#L1810

xuzhao9:
We are handling exceptions inside `_do_bench`: https://github.com/meta-pytorch/tritonbench/blob/161c9033ceee2609eadba1025e2385ad96a71841/tritonbench/utils/triton_op.py#L1809-L1816


There is an existing option `--keep-going` and we should reuse that.

PR #528, opened 2025-10-08
Enable backward mode for softmax operator

karthickai:
Stacked PRs:
 * __->__#528

--- --- ---


tritonbench/operators/softmax/operator.py
```diff
@@ -153,8 +283,17 @@ def get_input_iter(self):
             if additional_shapes:
                 shapes.extend(additional_shapes)
 
+        requires_grad = self.mode in (Mode.BWD, Mode.FWD_BWD)
```
xuzhao9: We can simplify this: https://github.com/meta-pytorch/tritonbench/blob/main/tritonbench/utils/triton_op.py#L743

PR #529, opened 2025-10-08
[Blackwell_Attention] [Triton] Make N_CTX const in DP FA kernel

njriasan:
Adds logic to make N_CTX const in both gluon and Triton for a fair comparison.

### DIFF
--- a/tritonbench/operators/blackwell_attentions/operator.py
+++ b/tritonbench/operators/blackwell_attentions/operator.py
@@ -28,7 +28,7 @@ try:
     )
 
     HAS_BLACKWELL_AUTOWS = True
-except (ImportError, IOError, AttributeError):
+except (ImportError, IOError, AttributeError, TypeError):
     # Needs compiler that supports autoWS
     HAS_BLACKWELL_AUTOWS = False
 


njriasan:
This seems to break on AutoWS for small sequence lengths. I suspect when N_CTX == BLOCK_N.

tritonbench/operators/blackwell_attentions/operator.py
```diff
@@ -470,7 +470,7 @@ def triton_tutorial_flash_dp_blackwell(
             q, k, v, self.causal, self.sm_scale, "ws"
         )
 
-    @register_benchmark(enabled=False)
```
neildhar, line 473: Can this actually be enabled in general? It looks like `HAS_BLACKWELL_AUTOWS` just checks whether the benchmarks exist, and not whether the compiler can support them.

tritonbench/kernels/gluon_attention_persistent_forward.py
```diff
@@ -211,7 +211,7 @@ class AttentionConfig:
     qk_scale: gl.tensor
     Z: gl.tensor
     H: gl.tensor
-    N_CTX: gl.tensor
+    N_CTX: gl.constexpr
```
neildhar: I wonder if we want to modify the gluon like this. Since the implementation is just copied from the examples, changes like this will probably end up getting overwritten. If it doesn't make a difference to performance, perhaps we can leave the gluon as is?

tritonbench/kernels/attention_utils.py
```diff
@@ -16,7 +16,7 @@
 PEEL_LAST = os.getenv("PEEL_LAST_ITER")
 WITH_TMA = os.getenv("WITH_TMA")
 HAS_EXPLICIT_WS = os.getenv("ENABLE_EXPLICIT_WS")
-SUPPORT_GLUON = os.getenv("WITH_GLUON")
+SUPPORT_GLUON = os.getenv("WITH_GLUON", "0") == "1"
```
neildhar: If we are immediately comparing with "1", why do we need the default value?

njriasan: Good point. It should just be not equal to None. Force of habit. Will update.

tritonbench/kernels/gluon_attention_persistent_forward.py
njriasan: Ya I'd be happy to revert the gluon changes. I just wanted to ensure we had a fair comparison and we weren't unfairly biasing AutoWS over gluon.

tritonbench/operators/blackwell_attentions/operator.py
njriasan: I think CI agrees with you. I will revert this.

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
```diff
@@ -253,6 +253,13 @@ def prune_invalid_configs(configs, named_args, **kwargs):
     return [conf for conf in configs if conf.kwargs.get("BLOCK_M", 0) <= N_CTX]
 
 
+def prune_persistent_configs(configs, named_args, **kwargs):
+    N_CTX = kwargs["N_CTX"]
+    # Filter out configs based on desired BLOCK_n
+    TARGET_BLOCK_N = 64 if N_CTX == 128 else 128
+    return [conf for conf in configs if conf.kwargs.get("BLOCK_N", 0) == TARGET_BLOCK_N]
```
njriasan, line 294: This will prune us to 1 config.

tritonbench/kernels/gluon_attention_persistent_forward.py
njriasan: Reverting based on feedback.

tritonbench/kernels/blackwell_triton_fused_attention_dp.py
manman-ren: minor: should we keep the condition from prune_invalid_configs?

njriasan: No. We will end up rejecting every config so the filtering won't work.

PR #536, opened 2025-10-09
Enable VECT_MUL for blackwell attention

neildhar:
Enabling `VECT_MUL` previously caused a regression, but this seems to come from the fma vectorisation in particular. Enable it just for multiplication for now, which seems to be a performance win.

### DIFF
--- a/tritonbench/kernels/blackwell_triton_fused_attention_dp.py
+++ b/tritonbench/kernels/blackwell_triton_fused_attention_dp.py
@@ -72,7 +72,8 @@ def _attn_fwd_subtile(
         qk -= m_ij[:, None]
     else:
         m_ij = tl.maximum(m_i, tl.max(qk, 1) * qk_scale)
-        if VECT_MUL:
+        # TODO: Figure out why vector FMA slows things down.
+        if VECT_MUL and False:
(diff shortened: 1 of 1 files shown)


tritonbench/kernels/blackwell_triton_fused_attention_dp.py
```diff
@@ -72,7 +72,8 @@ def _attn_fwd_subtile(
         qk -= m_ij[:, None]
     else:
         m_ij = tl.maximum(m_i, tl.max(qk, 1) * qk_scale)
-        if VECT_MUL:
+        # TODO: Figure out why vector FMA slows things down.
+        if VECT_MUL and False:
             qk = _fma_f32x2(qk, qk_scale, -m_ij[:, None])
```
manman-ren, line 77: Wondering if we should make VECT_MUL an integer with each bit representing one vectorization so we can autotune.

njriasan: I like that idea.

neildhar:
Superseded by https://github.com/meta-pytorch/tritonbench/pull/540

PR #540, opened 2025-10-09
[autoWS][FA] Add vectorization and fadd2_reduce

manman-ren:
Summary: mark N_CTX constexpr

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/tritonbench/kernels/blackwell_triton_fused_attention.py
+++ b/tritonbench/kernels/blackwell_triton_fused_attention.py
@@ -72,7 +72,7 @@ def _attn_fwd_subtile(
         qk -= m_ij[:, None]
     else:
         m_ij = tl.maximum(m_i, tl.max(qk, 1) * qk_scale)
-        if VECT_MUL == 2 or VECT_MUL == 3:
+        if VECT_MUL & 2:
             qk = _fma_f32x2(qk, qk_scale, -m_ij[:, None])
         else:
             qk = qk * qk_scale - m_ij[:, None]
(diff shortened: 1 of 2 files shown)


tritonbench/kernels/blackwell_triton_fused_attention.py
```diff
@@ -51,40 +70,58 @@ def _attn_fwd_subtile(
         qk -= m_ij[:, None]
     else:
         m_ij = tl.maximum(m_i, tl.max(qk, 1) * qk_scale)
-        qk = qk * qk_scale - m_ij[:, None]
+        if VECT_MUL == 2 or VECT_MUL == 3:
```
neildhar: Any reason not to test the bits directly?

```suggestion
        if VECT_MUL & 2:
```

tritonbench/kernels/blackwell_triton_fused_attention.py
```diff
@@ -51,40 +70,58 @@ def _attn_fwd_subtile(
...
     else:
         acc = acc * alpha[:, None]
 
+    PM: tl.constexpr = p.shape[0]
+    PN: tl.constexpr = p.shape[1]
+    if FADD2_REDUCE:
+        p0, p1 = p.reshape([PM, 2, PN // 2]).permute(0, 2, 1).split()
+        l_ij0, l_ij1 = tl.reduce((p0, p1), axis=1, combine_fn=_reduce_fadd2)
+        l_i0 = l_i0 * alpha + l_ij0
+        l_i1 = l_i1 * alpha + l_ij1
```
neildhar, line 107: Can we sum them after doing the reduction so we can keep the same interface? (and keep the differences localised to this part of the program)

Would that also save a register?

manman-ren: You mean "l_ij = l_ij0 + l_ij1" then a single l_i = l_i * alpha + l_ij? That is a good point.
The advantage is removing one addition inside the loop vs. register pressure of l_i1.
I basically copied from Gluon. And this is what is implemented in the dp version as well.
We can potentially clean this up if the alternative is better.

PR #567, opened 2025-10-15
Mamba2 Linear Attention Kernels

v0i0:


xuzhao9:
We need to add mamba_ssm install in the docker, but we can save it for later and assume user handles the installation themselves.

v0i0:
> We need to add mamba_ssm install in the docker, but we can save it for later and assume user handles the installation themselves.

Yeah mamba_ssm needs causal_conv1d which is a little tricky to install, so I'd leave it on the user end for now. if you lmk how to add stuff to docker, happy to give it a shot in a follow up!

PR #576, opened 2025-10-17
Add sweep inputs in flash_attention Tritonbench

sryap:
Summary: As title

Differential Revision: D84558841


xuzhao9 (approved):
Thanks for refactoring the inputs for flash_attention!

PR #582, opened 2025-10-21
gpda optimize the backward kernel

scxiao:
This PR include two changes:
- add the tuning config for the dot op
- optimization tanh computation using `fast_expf` and `fast_divf`

It can help the bwd kernel from 5.0ms to 2.8ms.

### DIFF
--- a/tritonbench/operators/gdpa/math.py
+++ b/tritonbench/operators/gdpa/math.py
@@ -59,8 +59,8 @@ def activation_string_to_int(s: str):
 
 @triton.jit
 def tanh(x):
-    exp_x = fast_expf(-2.0 * x)
-    return fast_dividef((1 - exp_x), (1 + exp_x))
+    exp_x2 = fast_expf(2.0 * x)
+    return fast_dividef((exp_x2 - 1), (1 + exp_x2))
 
 
 @triton.jit


xuzhao9:
Looks like related to https://github.com/meta-pytorch/tritonbench/pull/588, can you help take a look, @robieta ?

scxiao:
> Looks like related to #588, can you help take a look, @robieta ?

@robieta, I think the first part in my PR is the same as #588, seems like yours is cleaner, so you can merge you PR and I will do a rebase. I made some additional changes to the `tanh` here, could you please take a look? Thanks

tritonbench/operators/gdpa/math.py
```diff
@@ -59,8 +59,8 @@ def activation_string_to_int(s: str):
 
 @triton.jit
 def tanh(x):
-    # Tanh is just a scaled sigmoid
-    return 2 * tl.sigmoid(2 * x) - 1
+    exp_x2 = fast_expf(2.0 * x)
+    return fast_dividef((exp_x2 - 1), (1 + exp_x2))
```
robieta, line 63: Isn't this numerically unstable? (Since you quickly get to `inf / inf`)

scxiao: I saw this PR: https://github.com/meta-pytorch/tritonbench/pull/506/files changes it to `2 * fast_dividef(1.0, 1.0 + fast_expf(-2.0 * x)) - 1.0`, I think my changes are little simple. Do we have the assumption the `x > 0`?

scxiao: We are working on a `fast_tanh` in the Triton amd backend, I will wait that PR to see.

robieta: I just realized this is changing the non-approximate path. (There is already a path to use fastmath instead.) I did a pass on the ALU ops (https://github.com/meta-pytorch/tritonbench/pull/595). I apologize for the lack of visibility; internally we have diff stacks but I'm not sure if this repo supports something like ghstack.

scxiao: thanks @robieta for the info. I tried you PR and it has better performance than the changes in this PR, great changes. I will close this one.

PR #584, opened 2025-10-22
[power] Use power manager to plot gpu power and latencies

xuzhao9:
Add power manager to poll pynvml APIs during benchmarking.

The power manager runs in a child process and has two threads: the main thread and the data collecting thread. The main thread simply starts the data collecting thread and wait until the benchmark process finishes. The benchmarking process will spin up the power manager process when the benchmarking starts, and will send stop signal to the power manager process which will shutdown nvml and finalize the data as csv.

It also adds the `--repcnt ` CLI option to run a kernel with specified iterations. When used together with `--power-chart`, both power/frequency charts and latency charts will be saved.

Test plan:

```
python run.py --op gemm --only triton_tutorial_matmul --m 4096 --n 4096 --k 4096 --repcnt 10000 --power-chart --output-dir ~/power-chart
```

image

image

image

image

### DIFF
--- a/benchmarks/power_analysis/run.py
+++ b/benchmarks/power_analysis/run.py
@@ -39,6 +39,38 @@ from tritonbench.utils.run_utils import load_operator_by_args
 REPCNT = 2000
 
 
+workloads = [
+    # gemm
+    [
+        "--op",
+        "gemm",
+        "--only",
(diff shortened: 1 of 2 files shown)


adamomainz (approved):
This is great @xuzhao9 !!!

PR #592, opened 2025-10-27
[welford] Fix torch.compile impl to actually use Welford algorithm

yf225:


karthickai:
It would be great to include `layernorm` in the accuracy check https://github.com/meta-pytorch/tritonbench/blob/01829e16da92b43e16d6ef5ec611c8e0b2ad3474/tritonbench/operators/welford/operator.py#L127-L129 so we can verify that the eager Welford implementation is numerically correct.

yf225:
> It would be great to include `layernorm` in the accuracy check
> 
> https://github.com/meta-pytorch/tritonbench/blob/01829e16da92b43e16d6ef5ec611c8e0b2ad3474/tritonbench/operators/welford/operator.py#L127-L129
> 
> so we can verify that the eager Welford implementation is numerically correct.

Yeah it's slightly weird because we would then have another "real" baseline (i.e. layernorm) within the accuracy function itself. Maybe as a middle ground, we keep the eager impl to be layernorm, but just change the compile impl to use Welford algo.

karthickai:
> > It would be great to include `layernorm` in the accuracy check
> > https://github.com/meta-pytorch/tritonbench/blob/01829e16da92b43e16d6ef5ec611c8e0b2ad3474/tritonbench/operators/welford/operator.py#L127-L129
> > 
> > so we can verify that the eager Welford implementation is numerically correct.
> 
> Yeah it's slightly weird because we would then have another "real" baseline (i.e. layernorm) within the accuracy function itself. Maybe as a middle ground, we keep the eager impl to be layernorm, but just change the compile impl to use Welford algo.

that would work, then we can change name to `torch_compile_welford`

xuzhao9:
For accuracy checking, it is always available to use `--baseline <backend_name>` to specify a customized baseline. The one defined by `@register_benchmark(baseline=True)` is just the default baseline when not specified.

PR #608, opened 2025-10-31
Fix test failures

xuzhao9:
Summary: Fix most test failures.

Differential Revision: D85892639


adamomainz (approved):
thanks for fixing these!

PR #618, opened 2025-11-04
README: make install commands copy-pasteable and add missing `cd tritonbench`

fulvius31:


xuzhao9 (approved):
Thanks for the contribution 👍

PR #625, opened 2025-11-06
[tagging] Add tagging system to tags.yaml

xuzhao9:
Static analysis is always limited. For the ops that are still incorrect, users can manually add annotations.

We will add user tag annotations in the next PR.


FindHao (approved):
LGTM

PR #630, opened 2025-11-07
runner: override Helion path via TRITONBENCH_HELION_PATH

fulvius31:
Fallback to .install/helion when unset

Validate <helion>/benchmarks exists

add _env_get_str() helper

This will ensure flexibility.

### DIFF
--- a/tritonbench/utils/run_utils.py
+++ b/tritonbench/utils/run_utils.py
@@ -88,6 +88,7 @@ def get_run_env(
             run_env[f"{repo}_commit_time"] = "unknown"
     return run_env
 
+
 def _env_get_str(var_name: str, default: str) -> str:
     value = os.environ.get(var_name)
     if value is None:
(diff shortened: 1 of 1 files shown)


tritonbench/utils/run_utils.py
```diff
@@ -88,10 +88,28 @@ def get_run_env(
             run_env[f"{repo}_commit_time"] = "unknown"
     return run_env
 
+def _env_get_str(var_name: str, default: str) -> str:
+    value = os.environ.get(var_name)
```
xuzhao9, line 93: Shouldn't this be 
```suggestion
    value = os.environ.get(var_name, None)
```

fulvius31: Thanks for the review!

I think it's None by default already!

fulvius31:
@xuzhao9 is the pylint issue caused by this PR?

fulvius31:
@xuzhao9 I've just rebased since it looks like the pylint was fixed by another PR.

PR #649, opened 2025-11-18
Fix Matmul config issues

njriasan:
Summary:
Fixes 2 issues noted by our partners at Nvidia:

1. num_stages should be set as it controls the latency cost model parameter right now (I was wrong). It should be a value between 1-5.
2. The warp spec templates didn't propagate the num_ctas information.

Differential Revision: D87289927

### DIFF
--- a/tritonbench/operators/gemm/triton_matmul_configs.py
+++ b/tritonbench/operators/gemm/triton_matmul_configs.py
@@ -320,22 +320,25 @@ def get_tileir_configs():
     block_mnk_range = [16, 32, 64, 128, 256]
     occ_range = [1, 2]
     num_ctas_range = [1, 2]
+    num_stages_range = [1, 2, 3, 4, 5]
     configs = []
     for block_m in block_mnk_range:
         for block_n in block_mnk_range:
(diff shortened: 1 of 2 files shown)


tritonbench/operators/gemm/triton_matmul_configs.py
```diff
@@ -320,22 +320,25 @@ def get_tileir_configs():
...
-                            triton.Config(
-                                {
-                                    "BLOCK_M": block_m,
-                                    "BLOCK_N": block_n,
-                                    "BLOCK_K": block_k,
-                                    "GROUP_M": 8,
-                                    "occupancy": occ,
-                                },
-                                num_ctas=num_ctas,
+                        for num_stages in num_stages_range:
```
jananisriram, line 330: should we maybe set `num_warps` to a default here, or will triton handle that automatically?

njriasan: Triton automatically sets it to a default.

njriasan:
Failures appear to be unrelated. Merging.

PR #671, opened 2025-11-27
Your diff, D86682742, broke one test

xuzhao9:
Differential Revision: D87540514


xuzhao9:
Refactor Liger kernels to make them work on both internal and oss environments.

PR #672, opened 2025-11-28
Support alternative pip installers (uv) via get_pip_cmd

yuxuan-z19:
Currently, installation scripts always invoke pip via `sys.executable -m pip` in subprocesses.  
This PR introduces `get_pip_cmd()` (same as [PR#2659 for torchbench](https://github.com/pytorch/benchmark/pull/2659)), which:

- Uses the `PIP_MODULE` environment variable to override the installer if set.
    
    ```bash
    PIP_MODULE="uv pip" python install.py
    ```

- Defaults to `sys.executable -m pip` otherwise.

### DIFF
--- a/tools/python_utils.py
+++ b/tools/python_utils.py
@@ -40,8 +40,6 @@ def get_pkg_versions(packages: List[str]) -> Dict[str, str]:
 def get_pip_cmd():
     if env := os.getenv("PIP_MODULE"):
         return env.split()
-    if shutil.which("uv"):
-        return ["uv", "pip"]
     else:
         return [sys.executable, "-m", "pip"]
 


xuzhao9:
Thanks for the contribution. Unfortunately, our build environment is not ready for `uv` yet, even when `uv` is available. 

The main blocker is that we need conda env clone to construct and hot-switch multiple environments with the default pytorch triton, triton-main, and meta's triton fork. Right now, uv does not support multiple environments for single project.

We can accept the env `PIP_MODULE`, but we need to use `sys.executable -m pip` otherwise.

yuxuan-z19:
Got it @xuzhao9! I’ve removed the `uv` fallback, so the installer now only honors `PIP_MODULE` when provided and otherwise falls back to `sys.executable -m pip`. Could you take a look and let me know if this aligns with what you had in mind?

xuzhao9:
LGTM

PR #675, opened 2025-12-01
Support PID swizzling with causal FA

njriasan:
Summary: Adds PID swizzling to the causal example for FA. This boosts performance signifcantly (see the testing section).

Differential Revision: D88106034


htyu (approved):
Looks great!

PR #682, opened 2025-12-03
enable tma for output in gdpa kernel

Sibylau:
I was wondering why the GDPA kernel does not use TMA for storing the output.

This PR adds support for enabling TMA output storage in the GDPA kernel. 

Tested the performance and accuracy on H100 with 
```
bash denoise.sh python run.py --op gdpa --metrics speedup,accuracy --batch 1024 --max_seq_len 1024 --dim 512 --head 4 --sparsity 0.5 --only gdpa,gdpa_opt,gdpa_opt_sorted,eager_gdpa  --baseline eager_gdpa --simple-output --rep 3000
```
The original:
| gdpa-speedup  |  gdpa-accuracy   | gdpa_opt-speedup  |  gdpa_opt-accuracy  |  gdpa_opt_sorted-speedup    | gdpa_opt_sorted-accuracy |
|---|---|---|---|---|---|
| 73.3927 | 1 | 61.67 | 1 | 69.4485 | 1 |

After this PR:
| gdpa-speedup  |  gdpa-accuracy   | gdpa_opt-speedup  |  gdpa_opt-accuracy  |  gdpa_opt_sorted-speedup    | gdpa_opt_sorted-accuracy |
|---|---|---|---|---|---|
| 73.9545 | 1 | 60.1875 | 1 | 69.4483 | 1 |

### DIFF
--- a/tritonbench/operators/gdpa/gdpa.py
+++ b/tritonbench/operators/gdpa/gdpa.py
@@ -20,10 +20,8 @@ from typing import Tuple
 
 import torch
 import torch.nn.functional as F
-
 import triton  # @manual=//triton:triton
 import triton.language as tl  # @manual=//triton:triton
-
 from torch._library.triton import capture_triton
 from triton.tools.tensor_descriptor import TensorDescriptor
 
@@ -468,7 +466,8 @@ def _gdpa_fwd_compute(
                 [
                     (begin_q + start_m * BLOCK_M).to(tl.int32),
                     (o_offset).to(tl.int32),
-                ], acc.to(Out.type.element_ty),
+                ],
+                acc.to(Out.type.element_ty),
             )
         else:
             o_mask = (offs_m[:, None] < qlen) & (offs_d[None, :] < HEAD_DIM)


tritonbench/operators/gdpa/gdpa.py
```diff
@@ -158,7 +156,6 @@ def _gdpa_fwd_inner_ws(
             p = qk
 
         p *= qk_scale
-        p = p.to(v_dtype)
```
Sibylau, line 161: Deleted this type cast since it has a duplicate after the next if-else statement.

tritonbench/operators/gdpa/gdpa.py
```diff
@@ -20,10 +20,8 @@
 
 import torch
 import torch.nn.functional as F
-
 import triton  # @manual=//triton:triton
 import triton.language as tl  # @manual=//triton:triton
-
```
Sibylau, line 26: ufmt changes

njriasan:
> I was wondering why the GDPA kernel does not use TMA for storing the output.
> 

My experience in general has been that TMA store Isn't necessarily always beneficial/impactful in Triton. This is at least what I've seen with inductor integration at least. However, I don't fully know the reasons why.

That being said, this seems like a good change and seems to be net-neutral in perf.

njriasan (approved):

xuzhao9:
Feel free to ignore the MI350 test failure, it should be fixed by https://github.com/meta-pytorch/tritonbench/pull/686

PR #698, opened 2025-12-08
Add tflops metric to nightly benchmarks operators

leonematt:
This commit updates the nightly benchmark runs to include tflops metrics


xuzhao9:
Hi Matt! Thanks for the contribution.

Unfortunately, not all operators support the `tflops` metric because it will require users to implement the tflops formula like at https://github.com/meta-pytorch/tritonbench/blob/main/tritonbench/operators/gemm/operator.py#L516.

To make tflops metric available, we need to implement the formula for these operators.

leonematt:
@xuzhao9 I see, I will update this PR to bring those in :)

PR #705, opened 2025-12-10
[docker] speedup the nv docker build

xuzhao9:
Do not install CUDA compile toolchain in the docker to speedup the docker build.

Now the docker build finishes in < 20min, and it should reduce the chance of connection timeout when downloading the docker image.

Test plan:

https://github.com/meta-pytorch/tritonbench/actions/runs/20100057335


FindHao (approved):
LGTM. one question, are we using the docker's system cuda now?

xuzhao9:
@FindHao Previously we were using it to build fbgemm wheels (and for bisection, if required). We are not doing this anymore.

PR #710, opened 2025-12-11
Standardize IS_BLACKWELL definition

njriasan:
Summary: Standardizes the IS_BLACKWELL definition so kernels written for blackwell can be picked up on GB300.

Differential Revision: D88913502


xuzhao9 (approved):
Thanks for the BE! Can you help run `ufmt format .` to fix the linting error?

PR #711, opened 2025-12-11
Add static asserts to bwd

njriasan:
Summary: Adds static asserts to bwd for parameters that break the kernel when we tune them.

Differential Revision: D88964091


Sibylau:
maybe clear the CI by ufmt the file

PR #714, opened 2025-12-11
[ci] set path of libcublas.so

xuzhao9:
In https://github.com/meta-pytorch/tritonbench/pull/705, we removed the system cuda install and replace it with the cuda libraries bundled with torch nightly. However, the bundled libcublas.so couldn't be found which is required by certain operators.

Correctly setup `LD_LIBRARY_PATH` so that the bundled libcublas.so and libcublasLt.so can be found.

Test plan:

https://github.com/meta-pytorch/tritonbench/actions/runs/20174823071


xuzhao9:
@FindHao Yes, since we control the docker build, we always make sure there is a single ground of truth for the cublas library, and all cublas libraries across the system are consistent.

PR #722, opened 2025-12-12
Testing Module for Entropy Criterion and Online Linear Regression

jhou-jpg:
Summary:
This diff introduces a comprehensive testing module for verifying the accuracy of the Entropy Criterion and Online Linear Regression implementations. The ground truth for these tests are the naive, unoptimized variants of entropy and linear regression.

This diff should make it easier to plan future changes and improve dependability for these modules.

**Test Logic:**

Generate random data stream of length 600, for each element addition, verify that entropy / regression stats matches the unoptimized implementation

Datastreams: noisy (10 +- noise) , constant

Reviewed By: xuzhao9

Differential Revision: D89069384

### DIFF
--- a/test/test_gpu/accuracy_test/entropy_test_util.py
+++ b/test/test_gpu/accuracy_test/entropy_test_util.py
@@ -1,0 +1,321 @@
+"""
+Baseline utility functions for entropy criterion calculations.
+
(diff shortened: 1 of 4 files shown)


jhou-jpg:
seems like this exported the dependency diff as well, lmk if this is an issue and abandon

PR #725, opened 2025-12-13
[nightly] use profiler for latency on AMD

xuzhao9:
The AMD workflow is still noisy after pinning the runner: https://github.com/meta-pytorch/tritonbench/actions/runs/20187271035/job/57959493726

Try the following things:

1. print `amd-smi` in the nightly workflow to show if there is any other tenants
2. use profiler to minimize the host noise


xuzhao9:
We prefer to switch to disable dind.

PR #730, opened 2025-12-15
sync per critertion change

nmacchioni:
Summary: adding measurements to the entropy criterion is not free, but its work is entirely based on the CPU. we can be a bit faster by pipelining this work with the GPU work of the warmups actually running. the strategy is simple: we synchronize only on the event that we want to add to the entropy criterion, such that the subsequent GPU work can continue to process while the entropy criterion is updated

Reviewed By: jhou-jpg

Differential Revision: D89196043


jhou-jpg (approved):
looks good!

PR #739, opened 2025-12-17
[Gemm][CuTeDSL] Added Cutlass API benchmark

NikhilAPatel:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #741
* #740
* __->__ #739

Differential Revision: [D89393549](https://our.internmc.facebook.com/intern/diff/D89393549)

PR #740, opened 2025-12-17
[Gemm][CuTeDSL] Added Cutlass API exhaustive autotune benchmark

NikhilAPatel:
* #741
* __->__ #740
* #739

### DIFF
--- a/tritonbench/operators/gemm/operator.py
+++ b/tritonbench/operators/gemm/operator.py
@@ -526,7 +526,7 @@ class Operator(BenchmarkOperator):
 
             return out
 
-        @register_benchmark(enabled=HAS_CUTLASS_API)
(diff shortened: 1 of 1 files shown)


tritonbench/operators/gemm/operator.py
```diff
@@ -525,6 +526,24 @@ def out():
 
             return out
 
+        @register_benchmark(enabled=HAS_CUTLASS_API)
```
xuzhao9: How long does it take for exhaustive autotune? If it takes too long, maybe we want to set it to `enable=False` to avoid bloating the CI time. User can still force enable it with `--force`

NikhilAPatel: Yeah it takes quite a while I'll do that

PR #755, opened 2025-12-19
Add AK candidate conv1d in tritonbench

ying0129-meta:
Summary: Open source AK generated conv1d kernel

Differential Revision:
D89010513

Privacy Context Container: L1189098


tritonbench/operators/conv/__init__.py
```diff
@@ -0,0 +1,2 @@
+# (c) Meta Platforms, Inc. and affiliates. Confidential and proprietary.
```
xuzhao9, line 1: If this is an open-source operator, we shouldn't include this in the file header

PR #758, opened 2025-12-23
Add Diode Inductor max-autotune benchmarks to gemm, addmm, bmm

jananisriram:
Summary:
Add Inductor max-autotune benchmarking using Diode to TritonBench for the `gemm`, `addmm`, and `bmm` operators. This diff reduces friction in verifying Diode results against the base max-autotune results.

NOTE: The TritonBench benchmark currently includes the overhead of setting up and tearing down Diode within the benchmark, i.e. metrics like tflops and latency. Actual Diode benchmarking numbers should be slightly better than what TritonBench reports.

Differential Revision: D89398916

### DIFF
--- a/tritonbench/utils/diode_utils.py
+++ b/tritonbench/utils/diode_utils.py
@@ -1,0 +1,44 @@
+"""Diode (ML model for pruning autotuning configs) utils for TritonBench operators."""
+
+import diode.torch_diode.config as diode_config
+from diode.torch_diode.choices import DiodeInductorChoices
+from diode.torch_diode.models.triton_gemm.model import (
+    GEMMModelV2,
+    MODEL_CONFIGS,
+)
+from diode.torch_diode.registry import register, get_registry
+import logging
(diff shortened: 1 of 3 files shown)


NikhilAPatel (approved):

tritonbench/utils/diode_utils.py
```diff
@@ -0,0 +1,44 @@
+"""Diode (ML model for pruning autotuning configs) utils for TritonBench operators."""
+
+import diode.torch_diode.config as diode_config
```
NikhilAPatel, line 3: Might wanna try the `with try_import("HAS_DIODE"):` pattern here. https://github.com/meta-pytorch/tritonbench/blob/main/tritonbench/operators/gemm/operator.py#L34

PaulZhang12: Yeah second Nikhil here, diode will not always exist in OSS env and this will cause issues.

xuzhao9 (approved):

PR #759, opened 2025-12-27
Add Diode Inductor max-autotune benchmarks to gemm, addmm

jananisriram:
Summary: Add Inductor max-autotune benchmarking using Diode to TritonBench for the `gemm` and `addmm` operators. This diff reduces friction in verifying Diode results against the base max-autotune results.

Differential Revision: D89835679

### DIFF
--- a/tritonbench/operators/addmm/operator.py
+++ b/tritonbench/operators/addmm/operator.py
@@ -5,7 +5,7 @@ from typing import Any, Callable, Generator, List, Optional, Tuple
 import torch
 import torch._inductor.config as inductor_config
 import triton
-from tritonbench.utils.env_utils import is_fbcode
+from tritonbench.utils.env_utils import get_logger, is_fbcode
 from tritonbench.utils.python_utils import try_import
 
 try:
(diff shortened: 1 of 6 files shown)


tritonbench/operators/addmm/operator.py
```diff
@@ -149,6 +151,20 @@ def pt2_addmm_maxautotune(self, a, mat1, mat2) -> Callable:
             compiled(a, mat1, mat2)
         return lambda: compiled(a, mat1, mat2)
 
+    @register_benchmark(enabled=False)
+    def pt2_addmm_maxautotune_diode(self, a, mat1, mat2) -> Callable:
```
NikhilAPatel, line 155: Oh cool does this mean that Diode is enabled by default in max autotune mode now?

jananisriram: not yet, we're still testing Diode in production environments, but that's the goal for early 2026! the tritonbench patch just enables diode settings if "diode" is in the benchmark name

PR #770, opened 2026-01-08
[ci] pin meta triton commit

xuzhao9:
Update the meta triton commit.

Install nvidia-compute- package so libcuda.so.1 can be used by the test script.


FindHao (approved):
lgtm

PR #794, opened 2026-01-16
Arrive after loading m

njriasan:
Summary: Moves the arrive to after loading M. Based on the buffer definitions this is a safety requirement.

Differential Revision: D90878619


njriasan:
@htyu This was a divergence from the unit test. Not sure why this didn't get updated.

adamomainz (approved):

PR #809, opened 2026-01-21
Gate power manager import behind a GPU check

njriasan:
Summary: Report on the Triton slack that you can't use TritonBench without GPU software installed, which should be unnecessary if there is only the intention to benchmark GPU machines.

Differential Revision: D91147744

### DIFF
--- a/tritonbench/components/power/__init__.py
+++ b/tritonbench/components/power/__init__.py
@@ -1,1 +1,7 @@
-from .power_manager import PowerManagerTask
+import torch
+
+# PowerManagerTask requires pynvml which is only available when NVIDIA GPU is present
+if torch.cuda.is_available():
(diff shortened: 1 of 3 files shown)


njriasan:
@xuzhao9 Do we have a good environment to test this and see if there are other blockers?

njriasan:
Motivation: https://triton-lang.slack.com/archives/C01LY4FJL56/p1768987286043229

xuzhao9:
We expect users to run `python install.py` before running tritonbench, which includes pynvml as a hard dependency: https://github.com/meta-pytorch/tritonbench/blob/main/requirements.txt. Yes, there will be redundant packages installed if users only care about CPU or AMD.

Thanks for the PR. I will add a CPU-only test and think about how to handle this more elegantly.

PR #826, opened 2026-01-28
[CI] change docker build cadence to weekly

xuzhao9:
Due to the common breakage of the pytorch trunk with mslk and fbgemm_gpu, we do the following:

- build docker every week instead of nightly, during the weekend
- during test, do not do a fresh install of the environment. Instead, rely on the last successful built docker image.

Note that the nightly build will still run on the latest pytorch/mslk/fbgemm_gpu nightly.


xuzhao9:
The ROCM docker build failure is expected because of upstream pytorch issue: https://github.com/pytorch/pytorch/pull/173372

PR #845, opened 2026-02-03
Generalize custom benchmark with Diode-specific plugins

jananisriram:
Summary:
Generalize the previously Diode-specific custom benchmark in TritonBench in order to compare any two sets of benchmarks for a combination of operators and evaluation shapes.

`compare_benchmarks` compares 2 benchmarks (`--benchmarks-lhs`, `--benchmarks-rhs`) within an operator, scaled to multiple operators (`--ops`) on multiple sets of evaluation shapes (`--eval-shapes`).

`--custom-bench` allows the user to set custom benchmarking mappings. For example, the Diode custom benchmarking suite (`--custom-bench diode`) adds 2 command-line arguments and a custom set of benchmark mappings (between baseline max autotune and Diode max autotune).

Differential Revision: D92110855

### DIFF
--- a/benchmarks/compare_benchmarks/run.py
+++ b/benchmarks/compare_benchmarks/run.py
@@ -1,0 +1,415 @@
+"""
+Compare TritonBench benchmarks across operators and evaluation shapes.
+
+Usage:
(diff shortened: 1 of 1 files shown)


xuzhao9 (approved):
LGTM!

PR #853, opened 2026-02-04
add custom_shape_attention to tritonbench ops

sryap:
Summary: add an operator to support custom shapes in FA4

Differential Revision: D91758785


xuzhao9:
@sryap can you fix the linting error?

PR #871, opened 2026-02-16
Add CSV input support to addmm operator

romanmeta:
Summary:
The addmm operator's `--input` argument was declared in the arg parser but never wired into the shape selection logic. This adds a `read_shapes_from_csv()` function and connects `--input` to load shapes from a CSV file with columns M, K, N, Bias_1D_Y. This allows running all benchmark shapes in a single invocation instead of one per shape, eliminating repeated process startup overhead (~14x faster for 19 shapes).

Authored with Claude.

Differential Revision: D93372327


PaulZhang12:
LGTM! cc @xuzhao9

xuzhao9:
LGTM

PR #873, opened 2026-02-17
Adding a catch for triton's OutOfResources runtime error.

umechand-amd:
This PR is implementing a catch for Triton's Out of Resources runtime error just like torch.cuda.OutOfMemoryError to prevent crash and lead a graceful exit.


umechand-amd:
@xuzhao9 Thanks for approving this. The failures seem unrelated, can you please merge this?

PR #888, opened 2026-02-24
Fix TMA descriptor argument specialization error in Triton beta

dshi7:
Summary:
D92763334 reverted D92486154 and re-introduced native C++ argument specialization in Triton beta. The native specialization code (`specialize.cc`) handles known types (int, bool, float, tensor, TensorDescriptor, etc.) but does NOT handle TMA descriptor types (objects with `tma_desc_cpu_ptr` attribute). The old Python implementation had this handler, but it was removed in the native C++ rewrite.

The `_attn_fwd_base_opt` kernel uses configs (`configsOrig + configsOpt`) that all have `ENABLE_TMA: False`, meaning TMA descriptors are not actually used inside the kernel. However, the host code was unconditionally creating TMA descriptors when `HAS_TMA_DESC is True`, causing `TypeError: failed to specialize argument of type: KernelParamWrapper` when Triton beta tried to specialize these arguments.

The fix:
1. In `triton_fused_attention.py`: Only create `desc_helper` (TMA descriptors) for kernel variants that actually use TMA (`baseVariant != "base_opt"`). This avoids passing TMA descriptor objects to kernels that don't need them.

2. In `attention_utils.py`: Added defensive import of native `TmaDescKernelParam` from Triton to improve compatibility with TMA-enabled kernel variants in future.

Differential Revision: D94245198

### DIFF
--- a/tritonbench/kernels/attention_utils.py
+++ b/tritonbench/kernels/attention_utils.py
@@ -5,8 +5,6 @@ example and the proton implementation.
 
 import os
 
-import torch
-import triton
 import triton.language as tl
 
 # check if we have the TMA version in Triton PR #4498 (https://github.com/triton-lang/triton/pull/4498).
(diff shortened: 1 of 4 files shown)


xuzhao9:
`fp8_attention` and `gemm` still fails the H100 CI, cc @dshi7

xuzhao9:
We need to do the same here: https://github.com/meta-pytorch/tritonbench/blob/main/tritonbench/operators/gemm/persistent_matmul.py#L308

xuzhao9 (approved):

PR #899, opened 2026-02-27
[fp8_gemm_rowwise] Make fp8 config consistent with mslk

xuzhao9:
fp8 input generation depends on mslk. Directly use mslk utils to avoid breakage by mslk changes.

Test plan: CI


FindHao (approved):
lgtm

PR #939, opened 2026-03-10
Add inductor_flex_attention_fwd operator

OmarPavel:
Summary:
Add TritonBench operator to benchmark the flex attention forward pass
inductor kernel (triton_tem_fused_flex_attention_0).

Compares aten (eager flex_attention) vs inductor (torch.compile) with
causal mask, sweeping seq_len from 128 to 16384. Reports latency,
speedup, and tflops (adjusted for block sparsity).

Default config: B=8, H=16, D=128, bf16.

Reviewed By: stashuk-olek

Differential Revision: D95461825

### DIFF
--- a/tritonbench/operators/inductor_flex_attention_fwd/__init__.py
+++ b/tritonbench/operators/inductor_flex_attention_fwd/__init__.py
@@ -1,0 +1,1 @@
+from .operator import Operator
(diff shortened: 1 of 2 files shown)


xuzhao9:
I am curious why we are not using the existing `flex_attention` op? https://github.com/meta-pytorch/tritonbench/blob/main/tritonbench/operators/flex_attention/operator.py

Also, why we are creating separate ops for fwd and bwd pass? https://github.com/meta-pytorch/tritonbench/pull/940

PR #976, opened 2026-03-26
[FA][bwd] port bwd persistent implementation for autoWS

manman-ren:
Summary: With helion_release branch + https://github.com/facebookexperimental/triton/pull/1166, we are getting
```
                    (4, 32, 32, 8192, 8192, 128) bwd                                              937.674
```

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


xuzhao9 (approved):

manman-ren:
@xuzhao9 How do we import this now? Don't see the import button. Maybe we should also try using the tutorial version in the compiler, but this file also covers the tileIR test. Wondering if we should guard "attrs=, smem_budget=" which only exist in Meta's triton. CC @njriasan

PR #1002, opened 2026-04-09
Add Triton JIT launch latency profiling instrumentation and analysis report (#1002)

htyu:
Summary:

Add monkey-patch profiling to the launch_latency operator that breaks down
the per-call overhead of Triton's JITFunction.run into individual steps
(driver, binder, cache lookup, global check, grid, launch). Enable with
TRITON_LAUNCH_LATENCY_PROFILE=1. Also add analysis report comparing Triton
JIT dispatch (~11us) vs cuteDSL pre-compiled launch (~2.2us).

Reviewed By: tissue3

Differential Revision: D99636574

### DIFF
--- a/tritonbench/operators/launch_latency/operator.py
+++ b/tritonbench/operators/launch_latency/operator.py
@@ -1,3 +1,6 @@
+import os
+import time
+
 import torch
 from torch import zeros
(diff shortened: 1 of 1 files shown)


tissue3 (approved):
Review automatically exported from Phabricator review in Meta.

PR #1045, opened 2026-04-28
Fixing the MI350 segfaults

### LINKED ISSUE #1044: _do_bench_cudagraph_with_cache_clear segfaults on ROCm for BWD kernels with retain_graph=True after #1040
## Summary

After #1040, BWD-mode benchmarks (e.g. `rms_norm-bwd`, `layer_norm-bwd`) still segfault on AMD ROCm (gfx950/MI350X) when run with `--cudagraph --latency-measure-mode triton_do_bench`. The original `0f39f14` segfault was caused by huge `n_repeat`, but #1040's fix introduced a different problem: the new `estimate_cuda_runtime_ms` call inside `do_bench_wrapper` runs `fn()` **before** `_do_bench_cudagraph_with_cache_clear`, which corrupts GPU state for backward-pass functions and causes the subsequent HIP graph capture to segfault.

## Affected Kernels (Helion benchmark CI on MI350X)

- `rms_norm-bwd`
- `layer_norm-bwd`

These are the only `-bwd` kernels in Helion's CI. Other BWD benchmarks would likely also be affected.

## Reproduction

On AMD MI350X (ROCm 7.2):

\`\`\`bash
git clone https://github.com/pytorch/helion
cd helion
git clone https://github.com/meta-pytorch/tritonbench benchmarks/tritonbench
cd benchmarks/tritonbench
git checkout 0205245  # #1040
git submodule update --init --recursive
pip install -r requirements.txt
python install.py --liger
pip install -e . --no-deps
cd ../..

HIP_VISIBLE_DEVICES=0 python benchmarks/run.py \\
    --op rms_norm-bwd \\
    --metrics speedup,accuracy \\
    --latency-measure-mode triton_do_bench \\
    --cudagraph \\
    --only llama_rms \\
    --only-match-mode prefix-with-baseline \\
    --baseline llama_rms \\
    --num-inputs 1 \\
    --keep-going
\`\`\`

**Result:** Segmentation fault (exit code 139) immediately after \`Took Xms to get benchmark function for llama_rms\`.

## Root Cause

#1040 added this code at the top of \`do_bench_wrapper\` in \`tritonbench/components/do_bench/run.py\`:

\`\`\`python
if (warmup is None or rep is None) and not repcnt:
    estimate_runtime = estimate_cuda_runtime_ms(fn, grad_to_none=grad_to_none)
    warmup, rep = resolve_warmup_and_rep(warmup, rep, estimate_runtime)
\`\`\`

For BWD kernels, \`fn\` is the wrapped \`bwd_fn\` from \`BenchmarkOperator.get_bwd_fn()\`. On the first call, \`bwd_fn\`:

1. Runs the forward pass and stores the output in \`state[\"y\"]\`
2. Calls \`state[\"y\"].backward(state[\"dy\"], retain_graph=True)\`

When \`estimate_cuda_runtime_ms\` runs \`fn()\` 5+ times (estimation iterations) before \`_do_bench_cudagraph_with_cache_clear\`, the autograd state and GPU state get into a configuration that subsequently crashes when HIP graph capture begins.

**Empirical confirmation:** Skipping the outer \`estimate_cuda_runtime_ms\` call in \`do_bench_wrapper\` (passing \`rep=100\` directly) makes the BWD benchmark complete successfully on ROCm.

| Configuration | Result |
|---|---|
| \`aa5cb56\` (DEFAULT_REP=100, no outer pre-warm) | ✅ Completes (exit 0) |
| \`0f39f14\` (DEFAULT_REP=3000, no outer pre-warm) | ❌ SIGSEGV from huge n_repeat |
| \`0205245\` / #1040 (adaptive rep + outer pre-warm) | ❌ SIGSEGV — outer pre-warm corrupts BWD state |
| \`0205245\` with outer pre-warm bypassed | ✅ Completes (exit 0) |

## Proposed Fix

Avoid calling \`estimate_cuda_runtime_ms(fn)\` in \`do_bench_wrapper\` when \`use_cuda_graphs=True\`, since \`_do_bench_cudagraph_with_cache_clear\` already does its own internal estimation (5 GPU events) and applies \`resolve_warmup_and_rep\` again:

\`\`\`python
# In do_bench_wrapper:
if (warmup is None or rep is None) and not repcnt and not use_cuda_graphs:
    estimate_runtime = estimate_cuda_runtime_ms(fn, grad_to_none=grad_to_none)
    warmup, rep = resolve_warmup_and_rep(warmup, rep, estimate_runtime)
\`\`\`

This skips the redundant outer estimation for the cudagraph path while keeping #1040's adaptive behavior for non-cudagraph paths.

## Environment

- GPU: AMD Instinct MI350X (gfx950)
- ROCm: 7.2
- Triton: 3.6.0
- PyTorch: 2.10.0+rocm7.2

umechand-amd:
Fixes #1044
Fixes BWD-mode benchmark segfaults on AMD ROCm (gfx950/MI350X) introduced by #1040.                                 
   
  Problem                                                                                                             
                  
  #1040 added an outer estimate_cuda_runtime_ms(fn, grad_to_none=...) pre-warm in do_bench_wrapper. For BWD kernels   
  (which use retain_graph=True backward via BenchmarkOperator.get_bwd_fn), this pre-warm corrupts GPU state and causes
   the subsequent _do_bench_cudagraph_with_cache_clear HIP graph capture to segfault on ROCm.                         
                  
  Fix                                                                                                                 
  
  One-line change: skip the outer pre-warm when use_cuda_graphs=True. The cudagraph path already does its own internal
   5-event estimation and applies resolve_warmup_and_rep, so the outer call is redundant.
                                                                                                                      
  - if (warmup is None or rep is None) and not repcnt:                                                                
  + if (warmup is None or rep is None) and not repcnt and not use_cuda_graphs:
                                                                                                                      
  Validation (AMD MI350X, ROCm 7.2)                                                                                   
                                                                                                                      
  ┌────────────────┬────────┬─────────────────────────┐                                                               
  │     Kernel     │ Shapes │         Result          │
  ├────────────────┼────────┼─────────────────────────┤                                                               
  │ rms_norm-bwd   │ 5/5        │ ✅ pass (was: SIGSEGV)  │
  ├────────────────┼────────┼─────────────────────────┤                                                               
  │ layer_norm-bwd │ 10/10   │ ✅ pass (was: SIGSEGV)  │                                                               
  ├────────────────┼────────┼─────────────────────────┤                                                               
  │ rms_norm (FWD) │ 6/6        │ ✅ pass (no regression) │                                                               
  ├────────────────┼────────┼─────────────────────────┤                                                               
  │ softmax (FWD)  │ 19/19   │ ✅ pass (no regression) │
  └────────────────┴────────┴─────────────────────────┘

### DIFF
--- a/tritonbench/components/do_bench/run.py
+++ b/tritonbench/components/do_bench/run.py
@@ -660,7 +660,7 @@ def do_bench_wrapper(
         entropy_window_size: Size of rolling window for entropy tracking
         entropy_max_samples: Maximum samples before stopping warmup (safety limit)
     """
-    if (warmup is None or rep is None) and not repcnt:
+    if (warmup is None or rep is None) and not repcnt and not use_cuda_graphs:
         estimate_runtime = estimate_cuda_runtime_ms(fn, grad_to_none=grad_to_none)
         warmup, rep = resolve_warmup_and_rep(
             warmup,
--- a/tritonbench/components/do_bench/run.py
+++ b/tritonbench/components/do_bench/run.py
@@ -660,7 +660,7 @@ def do_bench_wrapper(
         entropy_window_size: Size of rolling window for entropy tracking
         entropy_max_samples: Maximum samples before stopping warmup (safety limit)
     """
-    if (warmup is None or rep is None) and not repcnt and not use_cuda_graphs:
+    if (warmup is None or rep is None) and not repcnt and latency_measure_mode == "triton_do_bench":
         estimate_runtime = estimate_cuda_runtime_ms(fn, grad_to_none=grad_to_none)
         warmup, rep = resolve_warmup_and_rep(
             warmup,
--- a/tritonbench/components/do_bench/run.py
+++ b/tritonbench/components/do_bench/run.py
@@ -660,7 +660,11 @@ def do_bench_wrapper(
         entropy_window_size: Size of rolling window for entropy tracking
         entropy_max_samples: Maximum samples before stopping warmup (safety limit)
     """
-    if (warmup is None or rep is None) and not repcnt and latency_measure_mode == "triton_do_bench":
+    if (
+        (warmup is None or rep is None)
+        and not repcnt
+        and latency_measure_mode == "triton_do_bench"
+    ):
         estimate_runtime = estimate_cuda_runtime_ms(fn, grad_to_none=grad_to_none)
         warmup, rep = resolve_warmup_and_rep(
             warmup,


xuzhao9:
Thanks for the fix! Can you take a look at the remaining errors?

umechand-amd:
> Thanks for the fix! Can you take a look at the remaining errors?

which errors? Can you point me to them?

xuzhao9:
Hi @umechand-amd , I see there are error tests in the unfinished CI:

image

tritonbench/components/do_bench/run.py
```diff
@@ -660,7 +660,7 @@ def do_bench_wrapper(
         entropy_window_size: Size of rolling window for entropy tracking
         entropy_max_samples: Maximum samples before stopping warmup (safety limit)
     """
-    if (warmup is None or rep is None) and not repcnt:
+    if (warmup is None or rep is None) and not repcnt and not use_cuda_graphs:
```
xuzhao9: Actually I think we can simply change this 
```suggestion
    if (warmup is None or rep is None) and not repcnt and latency_measure_mode == "triton_do_bench":
```

Because only `triton_do_bench` needs to run an extra round of run time estimation to determine warmup and rep. All other benchmarkers, we should switch to using `resolve_warmup_and_rep` to override the default warmup and reps.

What do you think?

umechand-amd: Yea that should work too.

umechand-amd:
> Hi @umechand-amd , I see there are error tests in the unfinished CI:
> 
> image

These are unrelated to this PR.

xuzhao9 (approved):
LGTM. Thanks!

PR #1089, opened 2026-05-27
Populate batch dimension for bmm, use direct shape columns (#1089)

jananisriram:
Summary:

Populate `batch` for bmm shapes (was hardcoded to `None`). Also read `m_dim`/`n_dim`/`k_dim` directly from `M`/`N`/`K` columns instead of re-parsing the shape string, falling back to `parse_shape()` for backward compatibility.

Reviewed By: warrendeng

Differential Revision: D106126315

### DIFF
--- a/benchmarks/compare_benchmarks/utils.py
+++ b/benchmarks/compare_benchmarks/utils.py
@@ -15,10 +15,13 @@ os.environ["TRITON_PRINT_AUTOTUNING"] = "1"
 import pandas as pd
 import torch
 import triton
-from dsi.logger.configs.TritonMultiOperatorBenchmarkComparisonsLoggerConfig.logger import (
-    TritonMultiOperatorBenchmarkComparisonsLogEntry,
-)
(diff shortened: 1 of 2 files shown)


benchmarks/compare_benchmarks/utils.py
```diff
@@ -181,19 +181,68 @@ def log_benchmark(
...
+            scale_a_recipe = None
+            scale_b_recipe = None
+            if (
+                op == "scaled_mm"
+                and scale_a_row is not None
+                and scale_a_col is not None
+                and scale_b_row is not None
+                and scale_b_col is not None
+            ):
+                from torch._inductor.fb.shape_logging import _scale_recipe
```
njriasan, line 212: Can this be reached in OSS? I think you need an is_fbcode() check.

jananisriram: good point, i've gated the imports at the top of the file in addition to this one. thanks!

warrendeng (approved):

PR #1090, opened 2026-05-27
Source eval shapes from inductor_mm_shapes Hive table (#1090)

jananisriram:
Summary:

Run Diode evals on the most frequently run subset of unique shapes (default 1500) from `inductor_mm_shapes`, a Hive table that logs shapes from all MAST jobs that use Inductor max-autotune. The previous implementation pulled shapes from `fbcode/pytorch/tritonbench/tritonbench/data/input_configs/`, which was 1. dependent upon human-authored, manual updates for more shapes and 2. required file path construction that was flaky and dependent upon a very specific file structure, which will not be maintained going forward.

Reviewed By: warrendeng

Differential Revision: D106441291

### DIFF
--- a/benchmarks/compare_benchmarks/run.py
+++ b/benchmarks/compare_benchmarks/run.py
@@ -1,15 +1,49 @@
 """
-Compare TritonBench benchmarks across operators and workloads.
+Compare TritonBench benchmarks across operators.
 
-Usage:
+Runs a pair of benchmarks (LHS vs RHS) on each operator and compares
+kernel selection and runtime. Results can be logged to the
+TritonMultiOperatorBenchmarkComparisons Scuba table.
+
+Shape source (priority order):
+  1. --input-loader <file.json>    Single JSON shape file.
+  2. --input-loader <dir/>         Directory of JSON shape files. Use
+                                   --input-filter to select by filename.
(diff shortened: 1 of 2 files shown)


benchmarks/compare_benchmarks/run.py
```diff
@@ -408,16 +391,35 @@ def parse_args(args: List[str] = None) -> BenchmarkConfig:
...
         default=None,
-        help="Substring filter on input config filenames (if searching a workload directory without explicit GPU subdirectories). Only files whose name contains this string will be used.",
+        help="Substring filter on filenames when --input-loader is a directory. "
+        "Only JSON files whose name contains this string will be used.",
+    )
+    parser.add_argument(
+        "--hive-job-filter",
+        type=str,
+        default=None,
+        help="Optional mast_job_name substring filter for Hive shape query. "
```
njriasan: Does this work in OSS? If its not usable outside of fbcode maybe it should be fbcode only.

jananisriram: great point, i've gated the fbcode-specific fields under `is_fbcode()`!

xuzhao9 (approved):
LGTM!

warrendeng (approved):

PR #1103, opened 2026-06-03
Deprecate legacy FP8 kernels

cthi:
Summary:
X-link: https://github.com/meta-pytorch/MSLK/pull/367

Remove:
- torch.ops.mslk.f8f8bf16_cublas
- torch.ops.mslk.f8f8bf16

FP8 Tensorwise is better served using `_scaled_mm` that would automatically use cublas/rocblas. We also remove f8f8bf16_cublas since it's only used in internal meta legacy code.

Differential Revision: D107286816


xuzhao9 (approved):
LGTM

PR #1105, opened 2026-06-05
tritonbench: add --helion-backend to override Helion backend

oulgen:
Summary:
Adds a generic `--helion-backend` CLI flag to tritonbench that overrides the Helion code-generation backend (e.g. `cute` for the CuTe DSL backend) for every operator's `helion` benchmark, without editing any operator or kernel files.

It works by (1) setting `HELION_BACKEND` so every `helion.kernel(...)` that does not pass an explicit backend picks it up (and subprocess children inherit it), and (2) wrapping `helion.kernel` to drop any hardcoded `config=`/`configs=` arguments when the backend is non-Triton. Those configs are Triton-specific (`indexing` / `pid_type` / `num_warps` / ...) and are not valid for other backends, so dropping them makes the kernel fall back to the chosen backend's default config.

Default behavior is unchanged (`triton`). Pair with `HELION_AUTOTUNE_EFFORT=none` to skip autotuning. The new logic lives in `tritonbench/utils/helion_utils.py` and is applied in `run_utils.py` right after arg parsing and at the start of `_run` (before the operator module is imported, so the wrap is active before any Helion kernel is built).

Reviewed By: mengluy0125

Differential Revision: D107679013

### DIFF
--- a/tritonbench/utils/helion_utils.py
+++ b/tritonbench/utils/helion_utils.py
@@ -1,0 +1,74 @@
+"""Generic Helion backend override for tritonbench.
+
+This lets you experiment with a non-default Helion code-generation backend
+(e.g. the CuTe DSL backend) for *any* operator's ``helion`` benchmark without
+editing the operator or kernel definitions.
+
+It works by:
+  1. Setting ``HELION_BACKEND`` so every ``helion.kernel(...)`` that does not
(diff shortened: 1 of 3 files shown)


mengluy0125 (approved):

oulgen:
Looks like tritonbench tests are failing on main so i will just land this

PR #1115, opened 2026-06-11
[AutoWS] [TritonBench] Remove pid computation for GEMM

njriasan:
Removes the PID recomputation that is used for the OAI FLATTEN cases.

authored with codex


jananisriram (approved):
lgtm!

PR #1122, opened 2026-06-15
[TPU] Add --device tpu option and torch_tpu init

norx1991:
First step toward running tritonbench's non-Triton baselines (eager ATen / PyTorch) on TPU via the out-of-tree `torch_tpu` backend. Plumbing only — no TPU timing path yet (follow-up).

- Adds `tpu` to the `--device` choices.
- Initializes the `torch_tpu` backend in `tritonbench_run` (explicit import + availability assert, so failures are loud and device init happens up front).

Synchronization needs no change: the per-benchmark sync already uses `torch.accelerator.synchronize()`, which was verified on TPU hardware to wait for all in-flight tensors on torch_tpu >= 2026-06-05 (the supported floor). CUDA behavior is unchanged.

### DIFF
--- a/tritonbench/utils/parser.py
+++ b/tritonbench/utils/parser.py
@@ -53,7 +53,7 @@ def get_parser(args=None):
         "--device",
         "-d",
         default="cuda",
-        choices=["cuda", "cpu", "mtia"],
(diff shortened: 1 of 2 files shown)


xuzhao9 (approved):
LGTM

PR #1129, opened 2026-06-16
[autoWS] support asymmetric blockwise mode for autows scaled_mm kernel

Sibylau:
## Summary
Refactor the autows Triton kernel to take per-operand scaling modes, and use that to support the DeepSeek-V3 asymmetric blockwise recipe (1x128 activations + 128x128 weights).

The previous autows kernel used a single `SCALING_MODE`, which treated A and B as one recipe. The operator, get_scale, torch `_scaled_mm_v2`, and vLLM are all per-operand. The kernel now takes `SCALE_A_MODE` and `SCALE_B_MODE` (real ScalingType values) and selects its algorithm from the pair:

| A mode | B mode | path |
|--------|--------|------|
| TensorWise | TensorWise | scalar scale in epilogue |
| RowWise | RowWise | per row/col scale in epilogue |
| BlockWise1x128 | BlockWise1x128 | MXFP8 via tl.dot_scaled (Blackwell) |
| BlockWise1x128 | BlockWise128x128 | DeepSeek in-loop fp32 rescale (new) |

DeepSeek scales change along K, so they cannot be applied in the epilogue. Each 128-wide K group's partial dot product is rescaled by `sa[m, kb] * sb[n//128, kb]` and then accumulated. Config pruning pins BLOCK_K=128 and BLOCK_N=128 for this path. 

## Test Plan
To test, drop `TRITON_USE_META_WS=1` as the autoWS pipeline is not supported yet. 

Accuracy (baseline: vllm_cutlass_fp8_gemm):
```
CUDA_VISIBLE_DEVICES=4 TRITON_ALWAYS_COMPILE=1 TRITON_KERNEL_DUMP=1 TRITON_DUMP_DIR=/tmp/triton_dumps  python run.py --op fp8_gemm   --scaling-pair BlockWise1x128,BlockWise128x128   --m 4096 --n 4096 --k 4096   --only vllm_cutlass_fp8_gemm,triton_autows_fp8_gemm   --metrics accuracy   --simple-output --baseline vllm_cutlass_fp8_gemm
```
shows
```
             x_val    triton_autows_fp8_gemm-accuracy
------------------  ---------------------------------
(4096, 4096, 4096)                                  1
               1.0
```
Performance:
```
CUDA_VISIBLE_DEVICES=4 TRITON_ALWAYS_COMPILE=1 TRITON_KERNEL_DUMP=1 TRITON_DUMP_DIR=/tmp/triton_dumps python run.py --op fp8_gemm --scaling-pair BlockWise1x128,BlockWise128x128 --m 4096 --n 4096 --k 4096 --only triton_autows_fp8_gemm,vllm_cutlass_fp8_gemm --metrics tflops
```
shows
```
             x_val    triton_autows_fp8_gemm-tflops    vllm_cutlass_fp8_gemm-tflops
------------------  -------------------------------  ------------------------------
(4096, 4096, 4096)                          813.287                         1594.27
           average                          813.287                         1594.27
```

### DIFF
--- a/tritonbench/operators/fp8_gemm/scaled_mm_autows.py
+++ b/tritonbench/operators/fp8_gemm/scaled_mm_autows.py
@@ -133,19 +133,22 @@ def _host_descriptor_pre_hook(nargs):
 
     sa = nargs.get("scale_a_ptr")
     sb = nargs.get("scale_b_ptr")
-    scaling_mode = nargs.get("SCALING_MODE", TENSORWISE)
-    if scaling_mode == ROWWISE:
+    a_mode = nargs.get("SCALE_A_MODE", TENSORWISE)
+    b_mode = nargs.get("SCALE_B_MODE", TENSORWISE)
+    if a_mode == ROWWISE and b_mode == ROWWISE:
         if isinstance(sa, TensorDescriptor):
             sa.block_shape = [block_m]
         if isinstance(sb, TensorDescriptor):
             sb.block_shape = [block_n // epilogue_subtile]
-    elif scaling_mode == BLOCKWISE_1x128:
+    elif a_mode == BLOCKWISE_1x128 and b_mode == BLOCKWISE_1x128:
         vec_size = nargs["VEC_SIZE"]
         rep_k = triton.cdiv(block_k // vec_size, 4)
         if isinstance(sa, TensorDescriptor):
             sa.block_shape = [1, block_m // 128, rep_k, 2, 256]
         if isinstance(sb, TensorDescriptor):
             sb.block_shape = [1, block_n // 128, rep_k, 2, 256]
+    # DeepSeek (BLOCKWISE_1x128 / BLOCKWISE_128x128) uses plain tensors, not
+    # TensorDescriptors, so there is nothing to resize here.
 
 
 def _get_autotune_configs():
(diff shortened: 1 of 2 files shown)


tritonbench/operators/fp8_gemm/scaled_mm_autows.py
```diff
@@ -198,10 +204,17 @@ def _prune_configs(configs, named_args, **kwargs):
             continue
         if N and config.kwargs["BLOCK_SIZE_N"] > N * 2:
             continue
-        # Block scaling requires BLOCK_K >= VEC_SIZE * 4
-        if scaling_mode == BLOCKWISE_1x128 and vec_size > 0:
+        # MXFP8 block scaling requires BLOCK_K >= VEC_SIZE * 4
+        if is_mxfp8 and vec_size > 0:
             if config.kwargs["BLOCK_SIZE_K"] < vec_size * 4:
                 continue
+        # DeepSeek blockwise (1x128/128x128): one 128-wide scale group per K
```
njriasan, line 211: Noob question: Can we reuse the tl.scaled_dot implementation at all here? Looking at the math I'm wondering if you can just use an identity to fuse A's scale into the hardware.

That way just the B scale could be in the epilogue.

njriasan (approved):
LGTM! One question regarding types of scaling but this makes sense overall.

tritonbench/operators/fp8_gemm/scaled_mm_autows.py
Sibylau: Great point; thanks for bringing it up. tldr is [vllm kernel explicitly requires the scalars to be fp32 type](https://github.com/vllm-project/vllm/blob/main/csrc/libtorch_stable/quantization/w8a8/cutlass/c3x/scaled_mm_helper.hpp#L15-L18) for fp8 scaled_mm kernels, so we cannot use tcgen05.mma.scaled (that only accepts ue8m0 for 32-K blocks or ue4m3 for 16-K blocks). 

The vllm has an parallel fp4 (e2m1) scaled_mm kernel which uses fp8 (ue4m3) as the scalar types and tcgen05.mma.scaled. We can focus on fp8 scaled_mm now and support fp4 scaled_mm as the next step. 

This autoWS triton kernel entangles two separate implementations into "BLOCKWISE_1x128, BLOCKWISE_1x128": one is MXFP8/MXFP8 (e8m0 scales, 1×32), and the other is BlockWise 1×128/1×128 (fp32 scales, per-128-along-K). Both has baselines so there are motivations to support both. I'm going to split into two separate kernels in the follow up PR.

PR #1131, opened 2026-06-18
fp8_gemm: split MXFP8 from fp32 BlockWise1x128 in autows kernel, add vLLM triton baseline

Sibylau:
## Summary

The previous autows kernel overloaded the (BlockWise1x128, BlockWise1x128) mode with MXFP8 (e8m0, 1x32 scale group), while the harness feeds that mode with fp32 1×128 scales. This caused the crash at this config with error message "RuntimeError: shape '[1, 32, 32, 2, 256]' is invalid for input of size 131072". (MXFP8,MXFP8) and (BlockWise1x128,BlockWise1x128) are two different scaling schemes; this PR splits them.

- (MXFP8,MXFP8) is now its own mode (ScalingType.BlockWise1x32): genuine 1×32 e8m0 microscaling, using tl.dot_scaled lowered to ttng.tc_gen5_mma_scaled. The harness "MXFP8" recipe routes to it. 
- (BlockWise1x128,BlockWise1x128) is now a symmetric fp32 path using tl.dot.
 
This PR also adds a new baseline vllm_triton_block_scaled_fp8_gemm (w8a8_triton_block_scaled_mm) — the only correct accuracy reference for symmetric (BlockWise1x128,BlockWise1x128), other baselines use cuBLAS _scaled_mm which returns CUBLAS_STATUS_NOT_SUPPORTED for it. 

## Test Plan
Accuracy for (BlockWise1x128,BlockWise1x128) passes:
```
CUDA_VISIBLE_DEVICES=4 TRITON_ALWAYS_COMPILE=1 TRITON_USE_META_WS=1 python run.py --op fp8_gemm --scaling-pair BlockWise1x128,BlockWise1x128 --m 4096 --n 4096 --k 4096 --only triton_autows_fp8_gemm,vllm_triton_block_scaled_fp8_gemm --metrics accuracy --simple-output --baseline vllm_triton_block_scaled_fp8_gemm
```
shows
```
             x_val    triton_autows_fp8_gemm-accuracy
------------------  ---------------------------------
(4096, 4096, 4096)                                  1
               1.0
```

Accuracy for (MXFP8,MXFP8) is blocked by compiler issue for tc_gen5_mma_scaled lowering. (TritonNvidiaGPUMMALoweringPass rejects the scale tmem_copy; the upstream 10-block-scaled-matmul.py tutorial fails identically):
```
CUDA_VISIBLE_DEVICES=4 TRITON_ALWAYS_COMPILE=1 python run.py --op fp8_gemm --scaling-pair MXFP8,MXFP8 --m 4096 --n 4096 --k 4096 --only triton_autows_fp8_gemm,torch_fp8_gemm --metrics accuracy --simple-output --baseline torch_fp8_gemm
```

### DIFF
--- a/tritonbench/operators/fp8_gemm/scaled_mm_autows.py
+++ b/tritonbench/operators/fp8_gemm/scaled_mm_autows.py
@@ -164,20 +164,17 @@ def _get_autotune_configs():
     configs = []
     block_configs = [
         # (BLOCK_M, BLOCK_N, BLOCK_K)
-        # (128, 128, 64),
-        # (128, 256, 64),
-        # (256, 128, 64),
-        # (128, 128, 128),
-        # (128, 256, 128),
+        (128, 128, 64),
+        (128, 256, 64),
+        (256, 128, 64),
+        (128, 128, 128),
+        (128, 256, 128),
         (256, 128, 128),
     ]
 
-    # num_stages=3 is needed by the symmetric fp32 1x128 path (the extra sb
-    # broadcast over M overflows SMEM at num_stages=4); other modes keep 4.
-    for num_stages in [3, 4]:
+    for num_stages in [3, 4, 5]:
         for BLOCK_M, BLOCK_N, BLOCK_K in block_configs:
-            for EPILOGUE_SUBTILE in [1]:
-                # for EPILOGUE_SUBTILE in [1, 2, 4]:
+            for EPILOGUE_SUBTILE in [1, 2, 4]:
                 configs.append(
                     triton.Config(
                         {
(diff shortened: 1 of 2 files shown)


tritonbench/operators/fp8_gemm/scaled_mm_autows.py
```diff
@@ -428,6 +464,40 @@ def scaled_mm_autows_kernel(
...
+                offs_k = ki * BLOCK_SIZE_K
+                a_tile = a_desc.load([offs_am, offs_k])
+                b_tile = b_desc.load([offs_bn, offs_k])
+                partial = tl.dot(
+                    a_tile,
+                    b_tile.T,
+                    out_dtype=tl.float32,
+                    allow_tf32=True,
+                )
+                sa = tl.load(
```
njriasan, line 490: Do we want to use TMA similar to the block scaled MMA if we have to load every iteration?

njriasan (approved):
LGTM! One minor comment but this makes sense.

tritonbench/operators/fp8_gemm/scaled_mm_autows.py
Sibylau: Thanks for bringing up this point. Right now the per-block scale loading seems not the bottleneck, as the scale_a/scale_b only consume a small amount of traffic. But there are gaps for loading, e.g., scale_a loading is not coalesced. I'm going to try optimizations in the following PRs.

PR #1148, opened 2026-06-24
[TPU] Pin a per-device default backend for torch.compile

norx1991:
torch_tpu autoloads on `import torch` and globally monkeypatches `torch.compile` to default to `backend="tpu"` (google-pytorch/torch_tpu#1912). On any machine where torch_tpu is installed this breaks `torch.compile` on CPU/CUDA tensors (`RuntimeError: ... tensor is expected to be on tpu, got cpu`), and it never sets the `dynamic=False` the TPU backend requires.

`set_compile_backend_for_device` (installed in `_run`, the per-op chokepoint all run paths funnel through) wraps `torch.compile` so calls without an explicit `backend=` get a device-appropriate default: `backend="tpu", dynamic=False` on TPU (Inductor-only `mode`/`options` kwargs are dropped, since the TPU backend rejects them), `backend="inductor"` otherwise. Calls passing an explicit backend are untouched, so the ~40 operator `torch.compile` call sites work per device with no per-operator edits. It is a no-op unless targeting TPU or torch_tpu is installed.

Verified on a TPU host (no automated TPU CI yet): `gemm` `pt2_triton_matmul` and `layer_norm` `torch_compile_layer_norm` now produce a `torch.compile` column on `--device tpu`; on CPU/CUDA `torch.compile` continues to route to Inductor.

### DIFF
--- a/tritonbench/utils/run_utils.py
+++ b/tritonbench/utils/run_utils.py
@@ -20,6 +20,7 @@ from tritonbench.operator_loader import get_op_loader_bench_cls_by_name, is_load
 from tritonbench.operators import load_opbench_by_name
 from tritonbench.operators_collection import list_operators_by_collection
 from tritonbench.utils.ab_test import compare_ab_results, run_ab_test
+from tritonbench.utils.compile_utils import set_compile_backend_for_device
 from tritonbench.utils.env_utils import (
     is_blackwell,
     is_cuda,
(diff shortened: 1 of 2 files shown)


tritonbench/utils/run_utils.py
```diff
@@ -659,6 +666,14 @@ def _run(args: argparse.Namespace, extra_args: List[str]) -> BenchmarkOperatorRe
         # Apply before the operator module is imported so the helion.kernel wrap
         # is in effect before any Helion kernel is constructed.
         apply_helion_backend_override(getattr(args, "helion_backend", None))
+        # Pin a device-appropriate default backend for torch.compile before any
+        # operator runs. Done here (the per-op chokepoint reached by every path:
```
xuzhao9: "chokepoint" -> "checkpoint"

norx1991: Thank you! Reworded the sentence to make it more clear.

xuzhao9 (approved):
LGTM

PR #1153, opened 2026-06-25
[AMD][CI] disable async copy to mitigate the segfault issue

xuzhao9:
As the title says


xuzhao9:
Setting the env won't work with the segfault

xuzhao9:
Setting this env won't work with segfault. Closing this.

PR #1158, opened 2026-06-26
[amd] re-enable segfault tests on AMD

xuzhao9:


xuzhao9:
They still have segfaults

PR #1166, opened 2026-07-03
[TPU] Make device-hardcoded operators/kernels honor --device

norx1991:
Four operators/kernels assumed CUDA and crashed at import or input generation under `--device tpu` (running Helion's benchmark bridge on TPU), which blocked their eager/Helion baselines from loading at all:

- `kernels/triton_fused_attention.py`: `is_cuda()` queried Triton's active-driver at import and raised on TPU (no active driver) — now returns `False` when there's no driver, fixing flash_attention + fp8_attention import.
- `operators/welford/`: the generated Triton file imported CUDA-only inductor symbols (`_cuda_getCurrentRawStream`, `_empty_strided_cuda`) absent on cpu/TPU torch builds; guarded them (used only by the Triton launch path) and gated the two Triton welford backends off when they're unavailable.
- input generators hardcoded the device (`welford` → `device="cuda:0"`, `low_mem_dropout` → `.cuda()`) — now use `self.device`.

Triton-only backends remain selected via `--only` as before (the established TPU strategy); this change only makes the operators loadable and their eager/Helion baselines runnable on TPU. On CUDA, behavior is unchanged (`self.device` is `cuda`; the guarded imports succeed).

Validated on a TPU v7 pod via the Helion bridge: flash_attention, fp8_attention, low_mem_dropout run; welford imports and reaches Pallas autotuning.

### DIFF
--- a/tritonbench/kernels/triton_fused_attention.py
+++ b/tritonbench/kernels/triton_fused_attention.py
@@ -38,7 +38,12 @@ except ImportError:
 
 
 def is_cuda():
-    return triton.runtime.driver.active.get_current_target().backend == "cuda"
+    try:
+        return triton.runtime.driver.active.get_current_target().backend == "cuda"
+    except (RuntimeError, AttributeError):
+        # No active Triton driver (e.g. on TPU, where these kernels never run) ->
+        # not CUDA. Avoids a hard crash at import time.
+        return False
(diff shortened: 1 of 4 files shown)


xuzhao9 (approved):
LGTM!

PR #1168, opened 2026-07-06
tritonbench grouped_gemm: add --shapes CLI flag for explicit G,M,N,K sweeps

robieta:
Summary: Adds a `--shapes` operator flag to the `grouped_gemm` op so production MoE shapes can be passed directly on the CLI instead of relying on the default/fb sweeps. Each entry is a `G,M,N,K` tuple (per group `x[M,K] @ w[K,N]`, G experts, M tokens/expert); the `M,N,K` ordering matches the SIR `bench_moe_grouped_gemm` shapes and TritonBench's other GEMM ops (e.g. `gemm` parses `MxNxK`). `parse_cli_shapes` expands each tuple into the split-size interface inputs (`W=[G,N,K]`, `split=[M]*G`) plus the 2D shared `B=[K,N]`, so both the split_size_* and `torch._grouped_mm`/triton backends run. `get_input_iter` consults `--shapes` first, overriding the default/fb sweep. Example: `--shapes 64,2048,3584,1792 64,2048,1792,1792`.

Differential Revision: D109464261

### DIFF
--- a/tritonbench/operators/grouped_gemm/operator.py
+++ b/tritonbench/operators/grouped_gemm/operator.py
@@ -88,6 +88,30 @@ def get_default_shapes():
     return shapes
 
 
+def parse_cli_shapes(specs):
+    """Build operator shape tuples from --shapes specs of the form 'G,M,N,K'.
+
+    Per group the GEMM is x[M, K] @ w[K, N]; G experts with M tokens each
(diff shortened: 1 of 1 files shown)


xuzhao9 (approved):
LGTM

PR #1172, opened 2026-07-06
[TPU] Make cross_entropy/embedding/fused_linear_jsd shape-configurable and TPU-dtype-safe

norx1991:
Enables `cross_entropy`, `embedding`, and `fused_linear_jsd` to run on TPU (via the Helion bridge / `--device tpu`), where tritonbench's large default shapes exceed TPU vmem and Pallas rejects int64 index tensors.

- **embedding**: add `--B/--T/--D/--v-range` shape args (defaults reproduce the existing two-shape × 8-vocab sweep) and generate int32 indices on TPU. **Now runs on TPU.**
- **cross_entropy**: generate int32 targets on TPU (Pallas rejects int64) and cast targets to Long in the torch baselines (`nn.CrossEntropyLoss` requires Long). This removes the int64 blocker; the Helion cross_entropy kernel then hits a separate Mosaic codegen limitation (Helion-side, tracked as a TODO), so it stays on `run_tpu.py` for now.
- **fused_linear_jsd**: add `--H/--V` shape args (defaults preserve `H=4096`/`V=128256`) so a TPU-sized vocab can be selected. The Helion example kernel still has a fp32-vs-bf16 `mm` dtype bug on TPU (Helion-side TODO), so it also stays on `run_tpu.py` for now.

int32 conversion is gated on `device == "tpu"` and all shape-arg defaults are unchanged, so CUDA behavior is identical. The shape args are also generally useful for benchmarking these ops at custom sizes on any device.

### DIFF
--- a/tritonbench/operators/cross_entropy/operator.py
+++ b/tritonbench/operators/cross_entropy/operator.py
@@ -53,14 +53,17 @@ class Operator(BenchmarkOperator):
                 requires_grad=True,
                 device=self.device,
             )
-            target = torch.randint(V, (self.B * self.T, 1), device=self.device).squeeze(
-                1
-            )
+            # Pallas/TPU does not support int64 tensors; use int32 indices there.
+            target_dtype = torch.int32 if self.device == "tpu" else torch.int64
+            target = torch.randint(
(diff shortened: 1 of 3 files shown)


xuzhao9 (approved):
LGTM

PR #1174, opened 2026-07-08
Fix scaled_mm eval coverage (#1160)

jananisriram:
Summary:

Fixes `scaled_mm` `compare_benchmarks` evals on Nvidia hardware:

- `diode/config.py`: allowlist `scaled_mm` (`aten::*`, `triton::*`)
- `compare_benchmarks/run.py`, `utils.py`: `--scaling-pair` passthrough
- `compare_benchmarks/fb/hive_shapes.py`, `run.py`: partition scaled_mm shapes by their real `(scale_x_recipe, scale_w_recipe)` from `inductor_mm_shapes` and thread the matching `--scaling-pair` through the eval, so shapes run under their real recipe instead of always tensor-wise
- `fp8_gemm.py`: pad block-scale construction so blockwise handles non-128-aligned shapes (Inductor/eager already use ceil_div; only the benchmark builder required exact divisibility)

Reviewed By: xuzhao9

Differential Revision: D109644490

### DIFF
--- a/benchmarks/compare_benchmarks/run.py
+++ b/benchmarks/compare_benchmarks/run.py
@@ -79,8 +79,14 @@ def build_op_args(
     config: BenchmarkConfig,
     benchmark_name: str,
     input_loader: Optional[str] = None,
+    scaling_pair: Optional[str] = None,
 ) -> List[str]:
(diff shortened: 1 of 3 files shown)


xuzhao9 (approved):
LGTM

PR #1192, opened 2026-07-20
[TPU] Fix _do_bench_tpu lazy-compile timing via torch_tpu pin (+ retain output)

norx1991:
**Update:** google-pytorch/torch_tpu#2494 has fixed the underlying torch_tpu issue (a device-wide sync now blocks on compiled-mode execution), so this PR updates tritonbench's `torch_tpu` pin to a build that includes the fix — replacing the earlier per-output materialize workaround.

---

Two measurement bugs in `_do_bench_tpu` (tritonbench's `--device tpu` wall-clock path):

1. **Output-drop.** The original timed a bare `fn()` then `synchronize()`, so the output was released before the sync and its `DeviceBuffer` teardown landed in the timed region. **Fix: bind the output across the sync (retain).** Pin-independent.
2. **Lazy `torch.compile` not blocked on.** On the torch_tpu build previously pinned, a device-wide `synchronize()` returned before a lazy `torch.compile` graph ran, so a compiled baseline read ~0 ms and its work inflated the next timed call (confirmed on-device: bare sync ~0.1 ms vs the real ~1.5 ms). This is **torch_tpu#2402**, fixed upstream by **torch_tpu#2494**. **Fix: bump `ci_commit_pins/torch_tpu.txt` to a build that includes it (`058f3a3e`)**, so `synchronize()` drains the work for every harness mode (fwd / bwd / no-grad).

The commit history shows the path: an earlier commit added a per-output `_materialize_output()` (mirroring `pytorch/helion#3111`); it is **replaced** here by the pin bump (mirroring the Helion-side decision — `pytorch/helion#3155` pin + `#3156` regression guard), because per-output materialize is fragile and incomplete — it can't force the work for `bwd`/`None`-returning harness modes, where the returned value isn't the timed work. Kept as a stacked commit (not squashed) so the switch is easy to review.

Note: the Helion bridge builds torch_tpu from **Helion's** pin, so the bridge picks up this fix via `pytorch/helion#3155`; tritonbench's own `torch_tpu.txt` governs tritonbench's standalone TPU path.


xuzhao9 (approved):

tritonbench/components/do_bench/run.py
```diff
@@ -489,15 +489,25 @@ def _do_bench_tpu(
...
+        t0 = time.time_ns()
+        _output = fn()
+        torch.accelerator.synchronize()
+        t1 = time.time_ns()
+        del _output
+        return (t1 - t0) * NS_TO_MS
+
+    # Warm up once, then estimate the per-call runtime.
+    _timed_call()
+    estimate_ms = sum(_timed_call() for _ in range(5)) / 5
```
xuzhao9, line 509: I am curious: does the first time of the `fn()` run include compile time? If it does, will averaging the `_timed_call()` time larger than expected kernel time, because the first run will include the compile time?

norx1991: Good question — there's a warmup call (line 509) right before the estimate with the result discarded, so the compilation will happen there and the estimate calls will reuse the compilation cache.

xuzhao9:
LGTM

PR #1198, opened 2026-07-24
[tlx] Guard torch_tlx_* providers on torch's tlx_mode inductor knob

wychi:
## Summary

The `torch_tlx_flex_attention`, `torch_tlx_addmm`, and `torch_tlx_mm` providers patch `triton.tlx_mode` via `inductor_config.patch(...)`, but were only gated on `has_tlx()` — which checks that the **Triton** TLX module exists, not that **torch's** inductor config exposes the `tlx_mode` knob.

On a torch build without `torch._inductor.config.triton.tlx_mode` (e.g. the MI350 CI's `torch 2.10.0+rocm7.0`), `inductor_config.patch()` raises `AttributeError` on `__enter__` while saving the prior value of the key — before `torch.compile` even runs — which fails the benchmark:

```
File ".../torch/utils/_config_module.py", line 870, in __enter__
    prior[key] = config.__getattr__(key)
AttributeError: torch._inductor.config.triton.tlx_mode does not exist
```

This was observed on MI350 CI for `flex_attention`; `addmm`/`gemm` carry the same latent bug (their `torch_tlx_*` providers use the identical pattern).

## Change

- Add `has_torch_tlx()` to `triton_utils.py` — the torch-side capability probe (`hasattr(torch._inductor.config.triton, "tlx_mode")`), the mirror of `has_tlx()` for the torch half of torchTLX.
- Require `has_torch_tlx()` on all three `torch_tlx_*` provider gates, so they self-disable when the knob is absent and run unchanged when it's present.

## Test

- `has_torch_tlx()` returns `False` on a torch without the knob (reproduces the CI condition) and `True` when present.
- Syntax + import verified on all four touched files.

🤖 Authored with Claude Code.

### DIFF
--- a/tritonbench/operators/addmm/operator.py
+++ b/tritonbench/operators/addmm/operator.py
@@ -13,7 +13,7 @@ from tritonbench.utils.env_utils import (
     is_hip_mi350,
 )
 from tritonbench.utils.python_utils import try_import
-from tritonbench.utils.triton_utils import has_tlx
+from tritonbench.utils.triton_utils import has_tlx, has_torch_tlx
 
 with try_import("HAS_HSTU"):
     try:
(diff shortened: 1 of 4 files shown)


xuzhao9 (approved):
LGTM!

PR #1207, opened 2026-07-30
Share one backward grad_output across implementations

norx1991:
The backward accuracy check generated the upstream gradient **per implementation** via `torch.manual_seed(0); dy = 0.1 * torch.randn_like(y)`, relying on `manual_seed` resetting the *device* RNG so every implementation draws the same `dy`. That holds on CUDA, but **not on torch_tpu at runtime** — there the device RNG isn't reliably reset by `manual_seed`, so each implementation drew a *different* `grad_output` and the backward accuracy check failed spuriously (e.g. `rms_norm-bwd`) even though the kernels were numerically correct (verified: grads match to ~1e-3 against an eager reference when driven with a shared gradient).

Fix: generate the `grad_output` **once** and share it across all implementations of an input (single-slot cache keyed by output shape/dtype/device), so the backward comparison feeds every implementation the identical upstream gradient — which is what a backward-accuracy comparison should do. Behavior-preserving on CUDA (same seeded values, just computed once); makes TPU behave the same.

Verified on TPU v7: `rms_norm-bwd` accuracy 0 → 1; forward kernels (e.g. gemm) unchanged.

### DIFF
--- a/tritonbench/utils/triton_op.py
+++ b/tritonbench/utils/triton_op.py
@@ -1294,6 +1294,10 @@ class BenchmarkOperator(metaclass=PostInitProcessor):
                 y_vals: Dict[str, BenchmarkOperatorMetrics] = functools.reduce(
                     _reduce_benchmarks, benchmarks, {}
                 )
+                # All impls of this input have run; drop the shared backward
+                # grad_output so it does not persist across inputs. The next input
+                # rebuilds it lazily on the first backward call (see get_bwd_fn).
+                self._shared_bwd_grad = None
                 metrics.append((x_val, y_vals))
                 logger.warning(
                     f"Completed input ID {input_id}:\n{table}",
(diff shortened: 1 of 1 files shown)


xuzhao9 (approved):
Interesting. Thanks for bring up the problem!

tritonbench/utils/triton_op.py
```diff
@@ -1373,8 +1373,31 @@ def bwd_fn():
...
+                # by (shape, dtype) is behavior-preserving on CUDA (same seeded
+                # values, just computed once) and makes TPU behave the same.
+                # Keep a single cached tensor (replaced when the output
+                # shape/dtype/device changes) rather than a growing dict, so a
+                # wide shape sweep does not retain one grad tensor per shape for
+                # the operator's lifetime. All implementations of a given input
+                # produce the same output shape and reuse this tensor; the next
+                # input replaces it.
+                y = state["y"]
+                grad_key = (tuple(y.shape), y.dtype, y.device)
```
xuzhao9: Can we explicitly cache the gradient tensors per-input, i.e., clean the cache after all impls of the current input is done, instead of keeping the cache alive across all inputs?

xuzhao9 (changes requested):
See inline comment. It would be better if we explicitly clear the cache after one input, instead of persisting the cache across all inputs.

tritonbench/utils/triton_op.py
norx1991: Great suggestion! Done.

xuzhao9 (approved):
LGTM

PR #1209, opened 2026-07-31
[TPU] Run welford at the requested --precision instead of hardcoded bf16

norx1991:
## What

`welford`'s `get_input_iter` hardcoded bf16 inputs and ignored `--precision`. Because `apply_precision` treats `fp32` as a no-op (it only casts for bf16/fp16), a caller could not actually run welford at fp32 — the inputs stayed bf16 regardless of the flag. This makes the operator honor `self.dtype`, defaulting to bf16 when precision is unset (`bypass` → `self.dtype` is `None`) so existing runs are unchanged.

## Why

At bf16 the layer-norm output drifts ~1 ULP on near-zero normalized values and exceeds the operator's own 1e-2 `assert_close` tolerance — for *every* correct implementation, including the pure-PyTorch reference welford compared against `F.layer_norm`. It's inherent bf16 rounding, not a kernel defect, so the accuracy metric reads 0 for all impls at bf16. Running the inputs at fp32 (which matches the canonical `examples/welford.py`) makes the comparison well-conditioned.

Manual verification on TPU (not covered by CI): the operator's accuracy flips 0→1 at `--precision fp32` and stays 0 at bf16.

The two CUDA Triton launchers also allocated their output buffer at hardcoded bf16; they now follow the input dtype, so `accuracy()` (which compares dtype) stays valid when a non-bf16 precision is requested.

### DIFF
--- a/tritonbench/operators/welford/operator.py
+++ b/tritonbench/operators/welford/operator.py
@@ -125,9 +125,13 @@ class Operator(BenchmarkOperator):
     def get_input_iter(self) -> Generator:
         for shape in self.shapes:
             s, d = shape
-            p1 = rand_strided((d,), (1,), device=self.device, dtype=torch.bfloat16)
-            p2 = rand_strided((d,), (1,), device=self.device, dtype=torch.bfloat16)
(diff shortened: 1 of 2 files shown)


xuzhao9 (approved):
LGTM!

PR #1214, opened 2026-08-04
Yield input tensors from get_input_iter for rope

Enna1:
rope was the only operator whose get_input_iter() yielded scalar (hidden_size, seq_length) tuples and constructed the actual q/k/cos/sin tensors inside each benchmark function. Every other operator yields its real input tensors from get_input_iter().


xuzhao9 (approved):
LGTM

PR #1221, opened 2026-08-05
[ci] fall back to fbgemm_utils when fbgemm is not installed

xuzhao9:
We disabled FBGEMM installation by default because of linking issues in nightly CI.

Re-implement some fbgemm util functions to make the CI happy.


jhou-jpg (approved):
lgtm!

PR #1224, opened 2026-08-10
[ci] Make aggregate metrics all-or-nothing

### LINKED ISSUE #753: Run "all-or-nothing" mode for avg metrics in the CI
For avg metrics, if the benchmark returns partial results, we should not return it because it will mislead users.

Similarly, when only partial results are available, set "pass" metric to 0.

Air-crypto:
## Summary
- suppress aggregate averages when a benchmark returns fewer inputs than requested or any backend reports an error
- emit the benchmark pass signal with the raw results and preserve it when CI maps metrics to the configured benchmark name
- add CPU regression coverage for complete, truncated, and failed benchmark results

Fixes #753

## Duplicate work
No open PR references #753 or implements all-or-nothing aggregate metrics as of August 10, 2026.

## Test plan
- [x] `.venv/bin/python -m unittest test.test_cpu.main.TestTritonbenchCpu.test_cpu_metric_x_only_true test.test_cpu.main.TestTritonbenchCpu.test_cpu_metric_custom_label test.test_cpu.main.TestTritonbenchCpu.test_cpu_list_operators_by_collection test.test_cpu.main.TestTritonbenchCpu.test_ci_aggregate_metrics_are_all_or_nothing test.test_cpu.main.TestTritonbenchCpu.test_ci_callback_preserves_partial_failure` (5 passed)
- [x] `.venv/bin/ufmt check tritonbench/utils/triton_op.py benchmarks/common.py test/test_cpu/main.py`
- [x] `git diff --check`
Full CPU module: the existing layer-norm test imports Triton, which is unavailable on macOS; the remaining five CPU tests pass locally

### DIFF
--- a/tritonbench/utils/triton_op.py
+++ b/tritonbench/utils/triton_op.py
@@ -320,7 +320,7 @@ class BenchmarkOperatorResult:
     simple_mode: bool
     # Tuple: (x_val, Dict[impl_name, BenchmarkOperatorMetrics])
     result: List[Tuple[Any, Dict[str, BenchmarkOperatorMetrics]]]
-    expected_num_inputs: Optional[int] = None
+    expected_num_inputs: int
     _result_dict: Optional[Dict[Number, Dict[str, BenchmarkOperatorMetrics]]] = None
 
     def _table(self):
@@ -364,10 +364,11 @@ class BenchmarkOperatorResult:
                 headers.append(f"{label}-{metric}")
         # generate rows
         hashes = {}
+        results = self.result
         if "kernel_source_hash" in self.metrics:
-            self.result.append(tuple(["hashes", {}]))
+            results = [*results, ("hashes", {})]
         avg_row = []
-        for x_val, y_val in self.result:
+        for x_val, y_val in results:
             col_num = 0
             row = []
             row.append(x_val)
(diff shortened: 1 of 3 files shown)


tritonbench/utils/triton_op.py
```diff
@@ -531,6 +532,22 @@ def userbenchmark_dict(self) -> Dict[str, Any]:
         # Userbenchmark Metric key format:
         # tritonbench_{op_name}_{op_mode}[{x_val}-{provider}-{metric}]
         userbenchmark_metrics_dict = {}
+        result_count = sum(1 for x_val, _ in self.result if x_val != "hashes")
+        expected_agg_values = (
+            self.expected_num_inputs
+            if self.expected_num_inputs is not None
```
xuzhao9: Can you help explain why `self.expected_num_inputs` could be None?

tritonbench/utils/triton_op.py
```diff
@@ -531,6 +532,22 @@ def userbenchmark_dict(self) -> Dict[str, Any]:
         # Userbenchmark Metric key format:
         # tritonbench_{op_name}_{op_mode}[{x_val}-{provider}-{metric}]
         userbenchmark_metrics_dict = {}
+        result_count = sum(1 for x_val, _ in self.result if x_val != "hashes")
```
xuzhao9: Can you help me understand why `x_val != "hashes"` is needed?

tritonbench/utils/triton_op.py
```diff
@@ -555,6 +572,8 @@ def userbenchmark_dict(self) -> Dict[str, Any]:
                         f"tritonbench_{benchmark_name}[x_{x_val}-{provider}]_{metrics}"
                     )
                     userbenchmark_metrics_dict[metric_name] = value
+                    if row_index >= result_count:
```
xuzhao9, line 580: I am not sure why this is needed. If it happens, should we raise Exception here instead?

tritonbench/utils/triton_op.py
```diff
@@ -570,8 +589,15 @@ def userbenchmark_dict(self) -> Dict[str, Any]:
                         agg_data[agg_metric_name] = agg_data.get(
                             agg_metric_name, []
                         ) + [numeric_value]
-        final_agg_data = {k: sum(v) / len(v) for k, v in agg_data.items()}
+        final_agg_data = {
+            key: sum(values) / len(values)
+            for key, values in agg_data.items()
+            if benchmark_complete and len(values) == expected_agg_values
+        }
```
xuzhao9: Can we have a warning message if it doesn't hold?

xuzhao9 (commented):
Thanks for your contribution! Please check out my inline comments.

tritonbench/utils/triton_op.py
Air-crypto: This handled a synthetic hash row that `_table` appended to `self.result`. I changed `_table` to use a local list, so the result count is now simply `len(self.result)`.

tritonbench/utils/triton_op.py
Air-crypto: It is required now. Every result gets `len(self._input_ids)` when it is constructed.

tritonbench/utils/triton_op.py
Air-crypto: These are expected display rows. `_table` adds hash and average rows after the input rows, so raising would break their JSON output. I added a short comment to make that clear.

tritonbench/utils/triton_op.py
Air-crypto: Added warnings for incomplete benchmark runs and per-metric value count mismatches. The regression test also checks the incomplete-input warning.

xuzhao9 (approved):
LGTM, thanks!!

Issue #270, opened 2025-07-09
[aten] Fix the broken PT2 ATen operators


xuzhao9:

xuzhao9:
Related issue: https://github.com/pytorch/pytorch/issues/158131

PR #1242, opened 2026-08-23
Support sparse COO inputs in the ATen loader

Air-crypto:
## Summary

- generate sparse COO tensors from serialized ATen input metadata instead of raising `NotImplementedError`
- preserve `sparse_dim` in newly serialized inputs while remaining compatible with the existing one-sparse-dimension DLRM payloads
- preserve shape, dtype, layout, nnz, dense dimensions, and coalesced state
- reject embedding inputs only when the requested operator is itself an embedding operator, so unrelated operators in the same input file remain loadable
- add CPU coverage for hybrid and pure COO tensors, legacy payloads, empty tensors, validation errors, and mixed embedding configs

Addresses the sparse-input portion of #270.

## Testing

- `python -m unittest test.test_cpu.test_aten_input_loader -v` (8 passed)
- adjacent CPU unit modules for device utilities, environment utilities, entropy comparison, and online regression (51 passed)
- loaded the repository's real `torchbench_train/dlrm_train.json` metadata for `aten.add_.Tensor` (15 payloads, including 6 sparse payloads)
- `ufmt check tritonbench/operator_loader/aten/input_loader.py test/test_cpu/test_aten_input_loader.py`
- `git diff --check`

The official `test.test_cpu.main` suite passed 5 of 6 tests locally. Its `layer_norm` test could not import because Triton does not provide a macOS wheel; the other five tests passed.

### DIFF
--- a/test/test_cpu/test_aten_input_loader.py
+++ b/test/test_cpu/test_aten_input_loader.py
@@ -1,0 +1,132 @@
+import unittest
+from types import SimpleNamespace
+
+import torch
+from tritonbench.operator_loader.aten.input_loader import (
+    deserialize_args,
+    deserialize_sparse_tensor,
+    OperatorInputLoader,
+    serialize_sparse_tensor,
+)
+
+
+class AtenInputLoaderTest(unittest.TestCase):
+    def test_hybrid_sparse_tensor_round_trip(self):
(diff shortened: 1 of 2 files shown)


xuzhao9 (approved):
LGTM. Thanks for the contribution!

PR #1249, opened 2026-08-27
[TLX][AMD] Benchmark optimized Flash Attention forward

bangtianliu:
Summary:

Reference FWD attention PR: https://github.com/facebookexperimental/triton/pull/2892

Use the landed optimized gfx950 cluster Flash Attention implementation for real TritonBench forward measurements.

This adds a separate `tlx_amd_fa_cluster` provider that:

- imports both `attention` and the landed `_validate_cluster_inputs` helper from `amd_fa_cluster`;
- is enabled only on MI350 when both landed symbols are available, preventing an older/partial tutorial module from being benchmarked accidentally;
- is marked forward-only, independently of backward tutorial availability;
- delegates input-contract checks to the tutorial validator during preprocessing, before benchmark timing; and
- preserves the landed dispatcher/autotuner by forwarding only Q, K, V, `sm_scale`, and `causal`, with no fixed launch configuration.

The existing `tlx_amd_fa_pipelined` forward/backward provider remains available for comparison.

The portable CPU tests cover TritonBench-owned behavior: forward-only registration and the complete `preproc -> validator -> attention` route, including call order and exact arguments. They use FakeTensor CUDA inputs. The validator and attention functions are recording stubs in that test because the default CPU-test Triton may predate #2892 and therefore cannot import the landed private validator; landed-Triton and MI350 runs exercise the real symbols.

## Performance

AMD MI350X, PyTorch `2.13.0+rocm7.1`, BF16, `B=4`, `H=48`, `D=64`. Each shape/mask pair was measured in an isolated TritonBench process. `torch.backends.cuda.preferred_rocm_fa_library()` reports `AOTriton`, so the SDPA column below is AOTriton.

| Mask | N | AOTriton SDPA (TFLOP/s) | TLX pipelined (TFLOP/s) | TLX cluster (TFLOP/s) | Cluster vs pipelined | Cluster vs AOTriton SNR* |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| non-causal | 512 | 207.955 | 344.148 | 355.152 | +3.2% | >= 40.56 dB |
| non-causal | 1024 | 316.425 | 517.877 | 524.411 | +1.3% | >= 40.56 dB |
| non-causal | 2048 | 566.801 | 589.895 | 607.558 | +3.0% | >= 40.56 dB |
| non-causal | 4096 | 631.724 | 647.213 | 653.305 | +0.9% | >= 40.56 dB |
| non-causal | 8192 | 673.171 | 675.044 | 684.174 | +1.4% | >= 40.56 dB |
| causal | 512 | 74.1185 | 137.424 | 142.532 | +3.7% | >= 40.56 dB |
| causal | 1024 | 150.417 | 223.539 | 308.987 | +38.2% | >= 40.56 dB |
| causal | 2048 | 258.213 | 271.946 | 437.886 | +61.0% | >= 40.56 dB |
| causal | 4096 | 359.008 | 302.229 | 543.949 | +80.0% | >= 40.56 dB |
| causal | 8192 | 586.615 | 313.780 | 619.024 | +97.3% | >= 40.56 dB |

*The minimum cluster-vs-AOTriton SNR across the sweep was 40.56 dB.

The cluster provider improves on the existing pipelined provider by 0.9%-3.2% for non-causal attention and 3.7%-97.3% for causal attention across this sweep.

Strict elementwise `allclose` can fail at long shapes because of rare AOTriton reference outliers rather than TLX corruption. For sampled worst rows at `N=8192`, the cluster output was substantially closer to a float32 reference:

- non-causal: AOTriton absolute error `3.2437e-2`; cluster absolute error `3.37e-5`;
- causal: AOTriton absolute error `9.6317e-2`; cluster absolute error `1.56e-4`.

The cluster provider intentionally leaves `config` unset. First use may run Triton autotuning; normal benchmark warmup absorbs it, while `--warmup 0` or first-call metrics such as peak memory can include tuning overhead.

## Test plan

- Default/stale-Triton focused CPU test: `2 passed`
- Landed-Triton focused CPU test: `2 passed`
- Full `test/test_cpu`: `96 passed`, plus 3 pre-existing dependency-pin failures
  - the failures expect `transformers==5.0.0rc3` while current `main` specifies `transformers==5.5.0`;
  - the same three failures reproduce on clean `main`
- Remaining CPU suite with the stale dependency-pin file excluded: `93 passed`
- Pinned `ufmt==2.9.1` on both changed files: clean
- Python `compileall` on both changed files: clean
- `git diff --check`: clean

### DIFF
--- a/test/test_cpu/test_flash_attention_tlx.py
+++ b/test/test_cpu/test_flash_attention_tlx.py
@@ -1,4 +1,3 @@
-import pytest
 import torch
 from torch._subclasses.fake_tensor import FakeTensorMode
 from tritonbench.operators.flash_attention import operator as flash_attention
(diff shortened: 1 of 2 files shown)


test/test_cpu/test_flash_attention_tlx.py
```diff
@@ -0,0 +1,152 @@
+import pytest
```
xuzhao9: May I ask why this is a CPU test? I saw there is `device="cuda"` specified in this test.

bangtianliu: This tests the TritonBench wrapper, not the GPU kernel: `fwd_only` registration, validation outside the timed region, and argument forwarding. `FakeTensorMode` makes `device="cuda"` metadata-only, so it runs without a GPU; real kernel correctness is covered on MI350.

bangtianliu:
cc @htyu for a look

xuzhao9 (approved):
LGTM

PR #1255, opened 2026-09-03
[TLX][AMD] Guard cluster Flash Attention regressions

bangtianliu:
## Summary

This PR adds dedicated regression coverage for the optimized gfx950 TLX cluster Flash Attention
forward provider.

- Check all 24 D128 FP16/BF16, causal/non-causal configurations from the optimization matrix.
- Add four D64 smoke configurations across both dtypes and mask modes.
- Require finite output, at least 35 dB SNR against SDPA, and sampled agreement with an explicit
  FP32 attention reference.
- Add a performance guard for BF16 `(B=1, H=64, N=16384, D=128, non-causal)`.
- Run the suite in a focused MI350 `meta-triton` CI job.

Both GPU tests compile in isolated child processes with
`DISABLE_LLVM_OPT=disable-machine-sink`. This matches the optimized configuration without leaking
the process-global LLVM option into unrelated tests.

## Performance guard

| Config | Reference | Local verification | Required floor |
| --- | ---: | ---: | ---: |
| `BF16, B=1, H=64, N=16384, D=128, non-causal` | 1010.5 TFLOP/s | 990.0 TFLOP/s | 925.0 TFLOP/s |

Timing uses the median of three independent windows, each with 500 ms warmup and 500 ms
measurement. The floor retains a clear separation from the 772.5 TFLOP/s pre-optimization result
while allowing for observed MI350 device variation.

## Test plan

- `python -m pytest -s --tb=short test/test_cpu/test_flash_attention_tlx.py` — 6 passed
- `python -m unittest test.test_gpu.test_flash_attention_tlx_regression -v` — 2 passed
  - numerical matrix: all 28 configurations passed
  - performance: 990.0 TFLOP/s
- Pinned `ufmt` check: passed
- Workflow YAML parsing: passed
- `bash -n .ci/tritonbench/test-gpu.sh`: passed
- `git diff --check upstream/main...HEAD`: passed

### DIFF
--- a/.ci/tritonbench/test-gpu.sh
+++ b/.ci/tritonbench/test-gpu.sh
@@ -23,4 +23,4 @@ export TORCHINDUCTOR_COMPILE_THREADS=1
 # best effort disable autotune to speedup test
 export TORCHINDUCTOR_MAX_AUTOTUNE=0
 
-python -m unittest test.test_gpu.main -v
+python -m unittest "${TEST_MODULE:-test.test_gpu.main}" -v
--- a/.github/workflows/_linux-test-mi350.yml
+++ b/.github/workflows/_linux-test-mi350.yml
@@ -7,6 +7,12 @@ on:
         type: string
         description: |
           Conda environment to activate when testing Triton
+      test_module:
+        required: False
+        type: string
+        default: "test.test_gpu.main"
+        description: |
+          Python unittest module to run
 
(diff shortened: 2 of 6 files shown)


bangtianliu:
cc @htyu  for a look

htyu (approved):
LGTM.

PR #1257, opened 2026-09-04
[ci] remove amd tlx test

xuzhao9:
The main tritonbench repo CI does not test against meta-triton by default as we have limited CI resources.

We should use https://github.com/facebookexperimental/triton repo to implement the TLX CI.

cc @bangtianliu

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -19,13 +19,6 @@ jobs:
     uses: ./.github/workflows/_linux-test-mi350.yml
     with:
       conda_env: "triton-main"
-  mi350-meta-triton-tlx-test:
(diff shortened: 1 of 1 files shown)


bangtianliu:
Thanks, @xuzhao9. Then, should I revert https://github.com/meta-pytorch/tritonbench/pull/1255/? And send it to the TLX side?

xuzhao9:
@bangtianliu This PR has already reverted the CI part and there is no need to revert the rest. You may submit another PR in https://github.com/facebookexperimental/triton to run `test_flash_attention_tlx_regression.py` (for example, adding it to https://github.com/facebookexperimental/triton/blob/main/.github/workflows/mi350.yml#L143).

PR #1264, opened 2026-09-14
Correct the compile_utils docstring for the fixed torch_tpu default

norx1991:
The module docstring said `torch_tpu` makes `backend="tpu"` the process-global default for `torch.compile`, so that merely installing it breaks `torch.compile` on cpu/cuda tensors. google-pytorch/torch_tpu#2639 changed that: the patched default now selects the TPU backend only when the inputs contain a TPU tensor.

Read today, the old text implies the `cpu`/`cuda` branch is what the module is for, when that branch is now compatibility for installations predating the fix. The TPU branch is the part carrying its weight, and for reasons the docstring never gave: the patched `torch.compile` still does not set the `dynamic=False` the TPU backend requires, and that backend rejects the Inductor-only `mode` / `options` kwargs many operators pass.

Documentation only — the executable AST is unchanged.

### Deliberately not changed

The `cpu`/`cuda` → `backend="inductor"` branch stays. tritonbench is paired with whatever torch_tpu a user has installed, not one pin, so it still protects anyone on a pre-#2639 build.

The docstring no longer claims pinning Inductor is equivalent to torch's own default. It isn't necessarily: an explicit `backend="inductor"` bypasses torch's default-backend selection, which can resolve to something else depending on version and config. Whether that override is still wanted on cpu/cuda now that #2639 has landed is a behavioral question, left alone here.


xuzhao9 (approved):
LGTM

PR #1265, opened 2026-09-14
Fix best_config metric with native autotuner (#1265)

xuzhao9:
Summary:

Fix best_config metric when using the native autotuner

Differential Revision: D120009655


agron911 (approved):
Review automatically exported from Phabricator review in Meta.

PR #1266, opened 2026-09-14
layer_norm tutorial: barrier before releasing the dw/db lock

mark14wu:
### Summary

`_layer_norm_bwd_dx_fused` releases the per-group `dw`/`db` lock with a scalar `tl.atomic_xchg(Lock, 0)` right after the `tl.store` of the partial sums. The release is issued by a single thread while the other warps' stores can still be in flight, so the next holder of the lock reads stale partial sums and overwrites them. The upstream Triton tutorial this file was copied from (`python/tutorials/05-layer-norm.py`) carries a `tl.debug_barrier()` between the stores and the release for exactly this reason; this PR adds it back. With `M = 8192, N = 2048` the missing barrier corrupts `dw`/`db` in ~99 of 100 runs on an RTX 4090.

### Cause

```python
    tl.store(DW, partial_dw, mask=mask)
    tl.store(DB, partial_db, mask=mask)
    # Release the lock
    tl.atomic_xchg(Lock, 0)
```

In the emitted PTX the scalar `atomic_xchg` is executed by one thread (`@p atom.global.gpu.acq_rel.exch.b32`) with no `bar.sync` between the `st.global` of `DW`/`DB` from all warps and that exchange. `acq_rel` only orders the releasing thread's own prior stores. Warps of the same program that have not yet issued their `DW`/`DB` stores race with the loads and stores of the next program that acquires the lock.

### Reproducer

```python
"""Reproducer for the dw/db lock race in tritonbench/operators/layer_norm/tutorial.py.

Usage: python repro_tritonbench_layernorm_lock.py /path/to/tutorial.py

Runs the tutorial's fused layer norm backward repeatedly and counts how
often dw/db disagree with torch autograd. With the lock released before the
partial-sum stores of all warps have completed, the next lock holder reads
stale partial sums and the accumulated dw/db are wrong.
"""
import importlib.util
import sys

import torch


def load(path):
    spec = importlib.util.spec_from_file_location("tutorial_under_test", path)
    mod = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(mod)
    return mod


def main(path, runs=100):
    tut = load(path)
    torch.manual_seed(0)
    for M, N in [(4096, 1024), (8192, 2048)]:
        x = torch.randn(M, N, device="cuda", requires_grad=True)
        w = torch.randn(N, device="cuda", requires_grad=True)
        b = torch.randn(N, device="cuda", requires_grad=True)
        dy = torch.randn(M, N, device="cuda")
        y_ref = torch.nn.functional.layer_norm(x, (N,), w, b, 1e-5)
        gx, gw, gb = torch.autograd.grad(y_ref, (x, w, b), dy)
        bad_w = bad_b = bad_x = 0
        for _ in range(runs):
            y = tut.layer_norm(x, (N,), w, b, 1e-5)
            dx, dw, db = torch.autograd.grad(y, (x, w, b), dy)
            bad_x += not torch.allclose(dx, gx, rtol=1e-2, atol=1e-1)
            bad_w += not torch.allclose(dw, gw, rtol=1e-2, atol=1e-1)
            bad_b += not torch.allclose(db, gb, rtol=1e-2, atol=1e-1)
        print(f"M={M} N={N}: dx mismatches {bad_x}/{runs}, dw mismatches {bad_w}/{runs}, db mismatches {bad_b}/{runs}")


if __name__ == "__main__":
    main(sys.argv[1])
```

Output on `main` (RTX 4090 (sm_89), triton 3.6.0, torch 2.10.0+cu128):

```text
M=4096 N=1024: dx mismatches 0/100, dw mismatches 0/100, db mismatches 3/100
M=8192 N=2048: dx mismatches 0/100, dw mismatches 98/100, db mismatches 99/100
```

### Fix

Add `tl.debug_barrier()` between the partial-sum stores and the lock release (the same lines as the upstream tutorial).

### Verification

Same script on this branch:

```text
M=4096 N=1024: dx mismatches 0/100, dw mismatches 0/100, db mismatches 0/100
M=8192 N=2048: dx mismatches 0/100, dw mismatches 0/100, db mismatches 0/100
```


xuzhao9 (approved):
Thanks for fixing this!

PR #1268, opened 2026-09-15
[welford] Avoid unrolling the torch.compile reduction

yushangdi:
The Welford baseline loops over every column in Python. `torch.compile` unrolls this recurrence, so the BF16 `(262144, 8192)` case generated 99 kernels on MI350X with PyTorch 2.13 / Triton 3.7.1 and spent many minutes compiling and tuning.

Express the reduction with `torch.var_mean(..., correction=0, keepdim=True)` so Inductor can lower it as a parallel Welford reduction. The tested 8192-column case generates one fused Triton kernel using the Welford helpers. FP32 accumulation, population variance, epsilon, affine parameters, output shape, and output dtype are preserved.

The reduction order changes, so this is **not bitwise equivalent** to the sequential baseline and benchmark timings will change. On a full-size normal BF16 input, 0.0244% of outputs differed, all within `atol=rtol=1e-2`. Large-offset inputs (`1000 + 4 * randn`, rounded to BF16) can differ beyond that tolerance; on the first 256 rows, mean absolute error against an FP64 reference rounded to BF16 improved from approximately `2.0e-3` to `7.0e-8`. This is an observed accuracy improvement, not a universal numerical guarantee.

Validation on AMD MI350X, PyTorch `2.13.0+rocm7.2`, Triton `3.7.1`:

- Local numerical validation passed for eager and compiled execution with FP32/BF16 normal, large-offset, and constant inputs against an FP64 reference.
- TritonBench CLI passed accuracy and cudagraph timing for `(262144, 1024)`, `(262144, 3072)`, and `(262144, 8192)`; all three inputs completed in approximately 16 seconds.
- Black formatting checks passed for the operator; `git diff --check` passed. The configured ufmt tool was unavailable locally.

```bash
python run.py --op welford \
  --only eager_layer_norm,torch_compile_welford --baseline eager_layer_norm \
  --metrics latency,accuracy --precision bf16 --input-id 0,4,9 \
  --cudagraph --latency-measure-mode triton_do_bench --atol 1e-2 --rtol 1e-2
```


yushangdi:
about 3.85× faster at width 1024.

```
  Measured on MI350X, BF16 (262144, 1024), PyTorch 2.13, with cudagraph enabled:

   Compiled implementation                             Runtime    Speedup
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━  ━━━━━━━━━
   Original sequential Welford                        0.790 ms         1×
  ─────────────────────────────────────────────────  ──────────  ─────────
   Current PR: two-pass reduction                     0.205 ms      3.85×
  ─────────────────────────────────────────────────  ──────────  ─────────
   New version with persistent reductions disabled    0.247 ms      3.20×
```

yushangdi:
note that CI is also failing on main: https://github.com/meta-pytorch/tritonbench/actions/runs/35012655052/job/104528187922

xuzhao9:
The failed CI is because of problematic AMD CI runner. Feel free to bypass.

yushangdi:
close because looks like it doesn't always do the welford algorithm and makes the name a bit untrue... I guess people can just use the layer_norm baseline instead.

PR #1269, opened 2026-09-16
[xpu] Turn off cooldown for grouped_mm perf measurement.

etaf:
## Motivation

We observed poor grouped_mm benchmark performance on XPU and traced it to the cooldown mechanism designed for NVIDIA clock-throttle recovery. On XPU, an idle gap puts the GPU into a low-power state, causing the first sample of each replica to incur a wake-up cost (~0.9 ms vs. ~0.29 ms at steady state on Arc Pro B70). The cooldown therefore creates the very outlier it is intended to prevent, inflating the measured latency.

## Solution

Disable cooldown on XPU to measure continuously, while preserving the existing behavior for other backends.


etaf:
Hi, @xuzhao9 could you help review this PR when free? Thanks.

xuzhao9:
Interesting! LGTM

cc the op owner @robieta

xuzhao9 (approved):

etaf:
Hi, @xuzhao9 @robieta could you help land this PR ? 

We need to enable the Helion benchmark CI workflow on XPU, which is currently blocked by updating the Tritonbench commit pin to a commit following this PR. However, I noticed that the current main branch is failing the CI job TritonBench PR Test / mi350-triton-main-test / linux-test-mi350 (push). Does this failure impact or block our plans to update the commit pin?

xuzhao9:
No it is not related - I have merged the PR.

